PR #12, opened 2020-04-03
Add Linux udev support, to correctly select the port meadow is attached to.

logicethos:
This will query udev, for idVendor 2e6a  (Wilderness Labs) on Linux Platforms.
libusb on it's own, doesn't provide the path.


logicethos:
MeadowSerialDevice makes use of Serial and Socket which are both IDisposable

I have made MeadowSerialDevice IDisposable as well, to make sure resources are cleanly disposed of.

I have made both Serial and Socket private, which will result in some breaking changes.
To fix these, I have added:
IsConnected
CloseConnection
SendData

I have fixed these in the VS Mac plugin, which will be in a separate pull request.

logicethos:
Main MeadowUsbDevice changes:

Boot mode detection.
await AwaitRemovedDevice - wait for meadow to disconnect (reset/boot/unplugged)
await AwaitAddedDevice -  wait for meadow to become accessible

Issue #40, opened 2020-12-17
SerialPort problem in AttemptToReconnectToMeadow()


Dblm0:
Using `Windows 7 x64` with `usbser.sys 6.1.7601.17514(win7sp1_rtm.101119-1850)` driver I faced a problem with device reconnection after its reset.

The  `System.IO.FileNotFoundException` is thrown on attempt to close the serial port.

CLI output on --MonoEnable cmd
```
Opening port 'COM32'
Data: Mono enabled, will run app.exe
App: Initialize hardware...
ID: Meadow by Wilderness Labs, Model: F7Micro, MeadowOS Version: 0.4.2 (Dec  8 2020 10:09:54),
Processor: STM32F777IIK6, Processor Id: 37-00-3d-00-0d-51-38-32-37-35-36-30,
Serial Number: 307335783238, CoProcessor: ESP32, CoProcessor OS Version: 0.4.2.0
App: Cycle colors...
Data: Mono has been enabled. Restarting F7 Micro
Host Serial Reconnect
Exception: System.IO.FileNotFoundException: Unable to find the specified file.
   at System.IO.Ports.SerialStream.Dispose(Boolean disposing)
   at System.IO.Stream.Close()
   at System.IO.Ports.SerialPort.Dispose(Boolean disposing)
   at System.ComponentModel.Component.Dispose()
   at System.IO.Ports.SerialPort.Close()
   at MeadowCLI.DeviceManagement.MeadowSerialDevice.Initialize(Boolean listen) i
n C:\Users\Dmitriev\Desktop\C#Projects\Meadow.CLI\Meadow.CLI.Core\DeviceManageme
nt\MeadowSerialDevice.cs:line 87
   at MeadowCLI.DeviceManagement.MeadowSerialDevice.AttemptToReconnectToMeadow()
 in C:\Users\Dmitriev\Desktop\C#Projects\Meadow.CLI\Meadow.CLI.Core\DeviceManage
ment\MeadowSerialDevice.cs:line 429
   at MeadowCLI.DeviceManagement.MeadowSerialDevice.DataReceived(Object sender,
MeadowMessageEventArgs args) in C:\Users\Dmitriev\Desktop\C#Projects\Meadow.CLI\
Meadow.CLI.Core\DeviceManagement\MeadowSerialDevice.cs:line 365
   at MeadowCLI.Hcom.MeadowSerialDataProcessor.ParseAndProcessReceivedPacket(Byt
[...]

PR #41, opened 2020-12-17
Win7 reconnect fix

Dblm0:
An attempt to solve #40 

Closing serialport at the beginning of AttemptToReconnectToMeadow() solved the problem with reset handling.

There might be a better way to fix this, but by now it allows me to use upload procedure in VS2019 meadow extension

### DIFF
--- a/Meadow.CLI.Core/DeviceManagement/MeadowSerialDevice.cs
+++ b/Meadow.CLI.Core/DeviceManagement/MeadowSerialDevice.cs
@@ -421,6 +421,9 @@ namespace MeadowCLI.DeviceManagement
 
         internal bool AttemptToReconnectToMeadow()
         {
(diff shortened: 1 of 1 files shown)


lamebrain:
hi @Dblm0 - I released an updated version of the VS extension that uses System.IO.Ports 5 which is more stable. Can you try that to see if it resolves your issue?

Dblm0:
> 
> 
> hi @Dblm0 - I released an updated version of the VS extension that uses System.IO.Ports 5 which is more stable. Can you try that to see if it resolves your issue?

Unfortunately, `0.13.1` release didn't resolve related issue.

Dblm0:
Since `0.12.8` update this PR doesn't solve the referenced issue.

PR #42, opened 2020-12-19
Multiple write delete file

dgr1992:
Let me know if this change is okay. If so I will update the documentation so it matches with the new WriteFile and DeleteFile

### DIFF
--- a/Meadow.CLI/Program.cs
+++ b/Meadow.CLI/Program.cs
@@ -18,7 +18,7 @@ namespace MeadowCLI
 {
     class Program
     {
-        
+
         [Flags]
         enum CompletionBehavior
         {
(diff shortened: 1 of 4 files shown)


Meadow.CLI/Program.cs
```diff
@@ -153,168 +153,197 @@ static async Task<CompletionBehavior> ProcessHcom(Options options)
...
-                  }
-                  else
-                  {
-                      if (string.IsNullOrEmpty(options.TargetFileName))
-                      {
+                if (options.WriteFile.Any<string>())
+                {
+                    string[] parameters = options.WriteFile.ToArray<string>();
+
+                    string[] files = parameters[0].Split(',');
```
lamebrain, line 160: Consider adding `StringSplitOptions.RemoveEmptyEntries | StringSplitOptions.TrimEntries` option to handle white space and and trailing commas.

PR #47, opened 2021-01-02
cli auth login (win only)

lamebrain:
* prepare PKCE auth login url
* open browser to prompt for creds
* spin up HttpListener to receive callback
* authorize callback and get token
* save refresh token in cred store

_Still need to add cred store for OSX and linux_

### DIFF
--- a/Meadow.CLI.Core/Identity/IdentityManager.cs
+++ b/Meadow.CLI.Core/Identity/IdentityManager.cs
@@ -97,6 +97,11 @@ namespace Meadow.CLI.Core.Auth
             {
                 throw new NotSupportedException();
             }
+            else
+            {
+                Console.WriteLine("Unsupported OS detected.");
+                throw new NotSupportedException();
+            }
 
             if (!string.IsNullOrEmpty(refreshToken))
             {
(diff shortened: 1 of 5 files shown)


Meadow.CLI.Core/Identity/IdentityManager.cs
```diff
@@ -0,0 +1,226 @@
...
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                throw new NotSupportedException();
+            }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+            {
+                throw new NotSupportedException();
+            }
+
```
Peter-Moody, line 100: Even though there are only 3 OSPlatform options (I assume), it makes me wonder whenever I see `else if` ladder and no terminal `else`.

Peter-Moody (approved):
Merge away!

Meadow.CLI.Core/Identity/IdentityManager.cs
lamebrain: Yeah, I'll add the else

PR #68, opened 2021-06-08
Refactor app deploy from vsmac extension

Redth:
Currently each extension seems to employ its own logic for deploying the app.

Taking the latest develop branch logic from the VSMac extension, this would move the deployment logic into the CLI tooling from which all extensions could share the same logic.


adrianstevens:
Good news - Pete has centralized the deployment logic in the CLI. He's just finishing up a massive refactoring which should improve stability but also greatly simplify deploying and debugging :)

adrianstevens:
I think we've got all of this covered and caught up :) closing PR

PR #84, opened 2021-07-18
Use downloaded BCL to override a project's local assemblies.

alexischr:
For all libraries in our downloaded BCL bundle, choose to deploy those to Meadow overriding any other assembly binaries in nugets and the bin/ directory


adrianstevens:
@alexischr do you have time to merge/rebase to the latest develop branch? looks like your branch is a few commits behind

alexischr:
@adrianstevens Done but the filename variable semantics of WriteFileAsync have changed so more so than before this could use testing/a review

adrianstevens:
I read through the PR and this all looks great. As I mentioned in #Slack, I think we should setup the CLI to also download the BCLs and check for updates so we're not relying on the end-user to call `meadow download os`.

adrianstevens (approved):

Issue #95, opened 2021-10-08
Incorrect serial port detected


HakanL:
On my Lenovo Carbon X1 machine I have a serial port named `Intel(R) Active Management Technology - SOL (COM3)`
and when the code in `GetMeadowSerialPortsForWindows` executes the powershell script to get the serial ports, it outputs this data:
```
Port: USB Serial Device (COM5), Name: Meadow F7 Micro, DeviceID: USB\VID_2E6A&PID_0001\335C336B3036
Port: Intel(R) Active Management Technology - SOL (COM3), Name: PCI Serial Port, DeviceID: PCI\VEN_8086&DEV_A0E3&SUBSYS_22D517AA&REV_20\3&11583659&1&B3
```

The problem is that `Intel(R)` gets detected as a serial port, and then fails to open.

adrianstevens:
Merged and published - closing issue

PR #98, opened 2021-10-08
Update MeadowDeviceManager.cs to only detect serial ports starting with COM

HakanL:
Proposed fix/workaround for https://github.com/WildernessLabs/Meadow.CLI/issues/95


CartBlanche:
@nitz Do you have any objections to closing this PR?

nitz:
Nope; this should be peachy with the WMI queries! :)

PR #100, opened 2021-10-16
Add `GetMeadowSerialPortsForWindows()`.

nitz:
Functioning more like the macOS and Linux counterparts, `GetMeadowSerialPortsForWindows()` runs a `WMIC` query to find Meadow ports on the machine. It specifically looks in the `Win32_PnPEntity` object for a device that has the class `Ports`, and a device ID that has Wilderness Labs' USB VID of 2E6A.

This allows the skipping of the false positives that are returned through the original `SerialPort.GetPortNames()` call. However, that functionality has been kept as a fallback, in case the WMIC query fails or otherwise returns no results.

An implementation note: I used running an external process of `wmic.exe`. This is not ideal, because it does lead to needing to parse the stdout, which could change as windows does. However, the System.Management classes are not available in .NET Standard 2.0, so as long as this Library targets that framework, this might be the closest we can get.

image

### DIFF
--- a/Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
+++ b/Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Text.RegularExpressions;
 using Meadow.CLI.Core.Devices;
 using Meadow.CLI.Core.Exceptions;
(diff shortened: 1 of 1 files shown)


adrianstevens:
it looks like WMIC is deprecated - we'll revert the code in the next release 

https://winaero.com/microsoft-removed-the-wmic-tool-from-windows-11-dev-builds/

nitz:
Lmfao. Seems like just my luck. Shouldn't need to revert it though, I'll put together a PR replacing WMIC with the powershell variety. 

I guess I have some other scripts in my life to update too 🙃

PR #107, opened 2021-11-02
Update DownloadManager.cs

alexischr:
We plan to release the CLI ahead of the OS release.


alexischr:
@adrianstevens this needs to be merged before the 0.6 OS release.

adrianstevens:
Since we're releasing in sync with the OS we can leave this check in place - closing PR

PR #111, opened 2021-11-10
Replace `wmic` tooling for `GetSerialPortsWindows()` with PowerShell equivalent

nitz:
Reverts be318543c6d65e00772067bd07177d496065cedd, and replaces `wmic`… based tooling with PowerShell equivalent.

As noted in #100 by @adrianstevens, it seems Microsoft is deprecating the `wmic` tool, and pointing users towards using the PowerShell cmdlets. This replaces the invocation of `wmic` in `GetMeadowSerialPortsForWindows()` with an invocation of the cmdlet `Get-CmiInstance` (aliased `gcmi`).

The end result of this is the same as the `wmic` tooling, but hopefully should be supported by Microsoft/Windows going forward.

`gcmi` shipped with Powershell 3.0 in 2012, so it is available by default on all platforms Windows 8 and newer. As before, any failure in invoking the cmdlet will result in falling back to the `GetSerialPorts()` method.

### DIFF
--- a/Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
+++ b/Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
@@ -290,14 +290,120 @@ namespace Meadow.CLI.Core.DeviceManagement
         public static IList<string> GetMeadowSerialPortsForWindows(ILogger? logger = null)
         {
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) == false)
+            {
                 throw new PlatformNotSupportedException("This method is only supported on Windows");
+            }
 
-            var ports = SerialPort.GetPortNames();
+            logger ??= NullLogger.Instance;
 
-            //hack to skip COM1
-            ports = ports.Where((source, index) => source != "COM1").Distinct().ToArray();
+            try
+            {
+                // Meadow devices use VID 2E6A
+                const string MeadowPnpDeviceIDPrefix = @"USB\VID_2E6A";
 
-            return ports;
+                // The 'Name' field is where the COM port name is.
+                string PnpEntityFields = "Name";
+
+                // The name field looks like "USB Serial Device (COM7)",
+                // so we want the port name between the parens.
+                const string ComPortNamePattern = @"\((COM[\d]+)\)";
+
+                // Can't use System.Management.ManagementObjectSearcher, not supported in netstandard2.0,
+                // and neither is System.Management.Automation from the PowerShell SDK.
+                // so we are using the powershell cmdlet `Get-CmiInstance` (alias gcim)
+                // see: https://devblogs.microsoft.com/powershell/introduction-to-cim-cmdlets/
+
+                string queryArgument = @$"gcim Win32_PnPEntity " + // get all the Win32_PnPEntities
+                    @$"| where {{ $_.PNPClass -eq 'Ports' -and  $_.DeviceID -like '{MeadowPnpDeviceIDPrefix}*' }} " + // filter by ports with our device id
+                    @$"| select {PnpEntityFields}" + // select the properties we want
+                    $@"| fl "; // format as a list
+
+                ProcessStartInfo psi = new()
+                {
+                    FileName = "pwsh.exe",
+                    Arguments = $"-NoLogo -NoProfile -NonInteractive -Command {queryArgument}",
+                    UseShellExecute = false,
+                    RedirectStandardOutput = true,
+                    RedirectStandardError = true,
+                    CreateNoWindow = true,
+                    WindowStyle = ProcessWindowStyle.Hidden,
+                };
+
+                using Process? proc = Process.Start(psi);
+
+                if (proc is null)
+                {
+                    logger.LogWarning("Failed to start PowerShell process.");
+                    throw new ApplicationException();
+                }
+
+                bool exited = proc.WaitForExit(5000);
+
+                if (exited == false)
+                {
+                    logger.LogWarning("PowerShell invocation hung longer than allowed timeout.");
+                    throw new ApplicationException();
+                }
+
+                string? output = proc.StandardOutput.ReadToEnd()?.Trim();
+                string? error = proc.StandardError.ReadToEnd()?.Trim();
+
+                if (proc.ExitCode != 0)
+                {
+                    logger.LogWarning("Failed to find Meadow by gcim, Powershell ExitCode: {ExitCode}", proc.ExitCode);
+                    throw new ApplicationException($"PowerShell unexpected ExitCode: {proc.ExitCode}");
+                }
+                else if (string.IsNullOrWhiteSpace(error) == false)
+                {
+                    logger.LogWarning("Failed to find Meadow by gcim: {error}", error);
+                    throw new ApplicationException($"PowerShell Execution error: {error}");
+                }
+
+                string[] rows = output?.Split(new[] { '\r', '\n' }, StringSplitOptions.RemoveEmptyEntries) ?? Array.Empty<string>();
+
+                if (rows.Length == 0)
+                {
+                    // if we didn't get any rows, fall back to regular port scanning.
+                    // tbd: is this desired behavior?
+                    throw new ApplicationException();
+                }
+
+                List<string> ports = new();
+
+                // go through all the result rows
+                foreach (string queryResult in rows)
+                {
+                    // pull the port name out of the string.
+                    MatchCollection? matches = Regex.Matches(queryResult, ComPortNamePattern, RegexOptions.IgnoreCase);
+                    if (matches.Count == 0 || matches[0].Success == false)
+                    {
+                        logger.LogInformation("Failed to extract port name from device entry: {device}", queryResult);
+                        continue;
+                    }
+
+                    // there is only one match, and our capture group is number 1. (group 0 is the whole match)
+                    string portName = matches[0].Groups[1].Value;
+                    ports.Add(portName);
+                }
+
+                return ports;
+            }
+            catch (ApplicationException aex)
+            {
+                // eat it for now
+                logger.LogDebug(
+                    aex,
+                    "This error can be safely ignored.");
+
+                // we failed to find a port using cmi, fall back to
+                // returning every potential port as reported by SerialPort.
+                string[]? ports = SerialPort.GetPortNames();
+
+                //hack to skip COM1
+                ports = ports.Where((source, index) => source != "COM1").Distinct().ToArray();
+
+                return ports;
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


CartBlanche:
@nitz Thanks for the PR. Why should we switch back to using PowerShell? Have you had issues or seen errors using the `SerialPort.GetPortNames();` call?

nitz:
Depends on what you classify as an `issue` :)

`SerialPort.GetPortNames()` returns a list of _all_ the serial ports on windows, regardless of what they are or where they came from. Using the WMI tooling allows Windows to have a closer parity to the way the other `GetSerialPortsFor<Platform>()` methods work, in that they only return ports that are either Meadow or Wilderness Labs specific.

This is especially useful when not specifying the port name manually every time, as especially in my setups, my Meadow serial port number is much later in the list, and so therefore would never get automatically selected. Filtering out other ports like this ensures that we only get the devices we want.

Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
```diff
@@ -290,14 +290,120 @@ public static IList<string> GetMeadowSerialPortsForLinux(ILogger? logger = null)
...
+                    MatchCollection? matches = Regex.Matches(queryResult, ComPortNamePattern, RegexOptions.IgnoreCase);
+                    if (matches.Count == 0 || matches[0].Success == false)
+                    {
+                        logger.LogInformation("Failed to extract port name from device entry: {device}", queryResult);
+                        continue;
+                    }
+
+                    // there is only one match, and our capture group is number 1. (group 0 is the whole match)
+                    string portName = matches[0].Groups[1].Value;
+                    ports.Add(portName);
```
CartBlanche, line 386: As per my earlier comment with returning the DeviceID, I was thinking it would be useful for customers to know which ID is on which port. So as the return DeviceID looks something like `USB\VID_2E6A&PID_0001\347238773536` 
It would be great to parse this to retrieve just the DeviceID and add to the port list, so that customers are presented with `COM5 (347238773536) in the IDE.

Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
```diff
@@ -290,14 +290,120 @@ public static IList<string> GetMeadowSerialPortsForLinux(ILogger? logger = null)
...
+                    throw new ApplicationException();
+                }
+
+                List<string> ports = new();
+
+                // go through all the result rows
+                foreach (string queryResult in rows)
+                {
+                    // pull the port name out of the string.
+                    MatchCollection? matches = Regex.Matches(queryResult, ComPortNamePattern, RegexOptions.IgnoreCase);
```
CartBlanche, line 377: This certainly does the job. IIRC Regex is (or was) notoriously slow. Have you done any per testing to see how it compares to other methods? Again my thoughts are about pre-empting upcoming Extension changes where this might be called several time a seconds, sometimes.

CartBlanche (changes requested):
Thanks again for taking the time to send us a PR. It is greatly appreciated!
I have finally had time to test it and have added some in-line comments.

Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
nitz: It's funny you bring this up — my local copy actually had this (just for printing out while I was testing.) I'd think the whole family of `GetMeadowSerialPortFor<Platform>()` functions would need to return something more than just an enumerable list of strings that are intended to be the raw port names in a case like this. E.g.: Many of my projects where I deal with serial ports I typically return a small record that describes the device name, ID if applicable, and other tertiary information available.

Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
nitz: I haven't done any explicit testing, no. I do know that in [.NET 5.0](https://devblogs.microsoft.com/dotnet/regex-performance-improvements-in-net-5/), Regex got some considerable love from the .NET team, that increased its performance by almost half an order of magnitude in several instances.

That said: The Regex wouldn't at all be the expensive path in this code. Spawning powershell and waiting for the result is nearly infinitely more time consuming than parsing the results in any means — I'll follow up more on this in a dedicated comment reply.

Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
```diff
@@ -290,14 +290,120 @@ public static IList<string> GetMeadowSerialPortsForLinux(ILogger? logger = null)
...
+                // and neither is System.Management.Automation from the PowerShell SDK.
+                // so we are using the powershell cmdlet `Get-CmiInstance` (alias gcim)
+                // see: https://devblogs.microsoft.com/powershell/introduction-to-cim-cmdlets/
+
+                string queryArgument = @$"gcim Win32_PnPEntity " + // get all the Win32_PnPEntities
+                    @$"| where {{ $_.PNPClass -eq 'Ports' -and  $_.DeviceID -like '{MeadowPnpDeviceIDPrefix}*' }} " + // filter by ports with our device id
+                    @$"| select {PnpEntityFields}" + // select the properties we want
+                    $@"| fl "; // format as a list
+
+                ProcessStartInfo psi = new()
```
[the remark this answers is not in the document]
nitz: Sure, this absolutely could be cached, I was just matching the style used in `GetMeadowSerialPortsForOsx()` and `GetMeadowSerialPortsForLinux()`.

Meadow.CLI.Core/DeviceManagement/MeadowDeviceManager.cs
```diff
@@ -290,14 +290,120 @@ public static IList<string> GetMeadowSerialPortsForLinux(ILogger? logger = null)
...
+                // so we want the port name between the parens.
+                const string ComPortNamePattern = @"\((COM[\d]+)\)";
+
+                // Can't use System.Management.ManagementObjectSearcher, not supported in netstandard2.0,
+                // and neither is System.Management.Automation from the PowerShell SDK.
+                // so we are using the powershell cmdlet `Get-CmiInstance` (alias gcim)
+                // see: https://devblogs.microsoft.com/powershell/introduction-to-cim-cmdlets/
+
+                string queryArgument = @$"gcim Win32_PnPEntity " + // get all the Win32_PnPEntities
+                    @$"| where {{ $_.PNPClass -eq 'Ports' -and  $_.DeviceID -like '{MeadowPnpDeviceIDPrefix}*' }} " + // filter by ports with our device id
```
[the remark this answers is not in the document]
nitz: This didn't work as intended where? Were you trying to reproduce the results through powershell directly? In that case you actually only need a single `{` and `}`, the doubling there is for the string interpolation to emit a single brace, rather than try and start an interpolation expression.

nitz:
With regards to [this](https://github.com/WildernessLabs/Meadow.CLI/pull/111#discussion_r768081001) post:

Let me preface first that when I authored this (and even the original #100), I did so with the intent to mimic and behave as closely as possible to the other supported platforms.

Running a powershell instance that invokes WMI tooling is a slow process. If there's the chance (or intent) that this code could be invoked multiple times per second, then we might need to take a look at the biggest goal for what this function would provide.

### Speed
If speed was the absolute goal, then the `SerialPort.GetPortNames()` is going to be the fastest method by far. However, while it's quick, it's effectively useless in terms of accuracy. In cases where Meadow is the only `COM` device on the user's machine, it would seem to work perfectly, but as soon as they have more than a single device, it results in "false positives" that would likely serve to confuse and frustrate users.

### Accuracy
The WMIC approach is always going to get up to date, accurate information. It's what would let us filter out ports/devices that we (and therefore the user) aren't interested in, but it comes at the cost of speed. The majority of the time spent would be in the startup, run, and teardown of the process that is actually running the WMIC query. The benefit of this approach is that it requires no additional direct dependencies. It is, however inherently "fragile" (I suppose in the same sense that the other platforms enumerate the `/dev` node is fragile.)

However, .NET 5+ includes the `System.Management` libraries that allow for direct invocation of WMIC from the dotnet runtime, which would remove the majority of the overhead that's associated with the current approach.

But: since we're only targeting .NET Standard 2.0, we don't have the ability to use `System.Management` directly, as it doesn't ship as part of the .NET runtime. It was split out into the `Microsoft.Windows.Compatibility` package, and could be included in `Meadow.CLI.Core` via it's [nuget Package](https://www.nuget.org/packages/System.Management/). This dependency would also include an implicit dependency on `System.CodeDom`.

### Conclusion

All of that said, I would highly advocate for the `System.Management` approach, but have always been hesitant to tell a project to add additional dependencies, especially for a dependency that would be platform specific. However, my advocacy would come from the point that it allows for the type safe execution of queries that would suffer from neither the potential fragility of the external process approach, and neither the overhead of needing to create and tear down an external process.

If it's something you're interested in seeing, it would take minimal time to convert the current changes to use it instead.

CartBlanche:
@nitz Thanks for your comments. We are still discussing internally the aspects of this change as it may impact those people using the command line tools too. Once we decide either way, I'll post here.

nitz:
The command line tool is actually what prompted the change (and #100) in the first place. I have several physical serial ports on my machines, and a few USB COM ports too, so my Meadow device would consistently get a COM port number towards the end of that list. This lead to the CLI tooling failing to automatically select the correct port any time I didn't specify `-s COM<X>` every time, which got to be quite thorny to remember. :)

I'll spin up what it looks like with `System.Management` (hopefully today,) so you can take a peek at that too.

CartBlanche:
Strangely enough, I was also playing with System.Management this morning, while waiting for the US/Canada to come online.
A Better version of this perhaps...

            ```
using (var searcher = new ManagementObjectSearcher ("SELECT * FROM Win32_PnPEntity WHERE PNPClass like 'Ports'"))
            {
                var query = searcher.Get();
                var ports = query?.Cast<ManagementBaseObject>().ToList();
                if (ports.Count > 0)
                {
                    foreach (var item in ports)
                    {
                        var totalDeviceID = item["DeviceID"] as string;
                        var caption = item["Caption"] as string;
                        if (totalDeviceID.Contains("USB\\VID_2E6A"))
                        {
                            var splitDevice = totalDeviceID.Split("\\");
                            var openBracketPosition = caption.IndexOf("(") + 1;
                            var closeBracketPosition = caption.IndexOf(")");
                            var stringLength = closeBracketPosition - openBracketPosition;
                            var parsedCOMPort = caption.Substring(openBracketPosition, stringLength);
                            Console.WriteLine("{0} ({1})", parsedCOMPort, splitDevice[2]); // Add this to the list
                        }
                    }
                }
                else
				{
                    Console.WriteLine("No Devices Found");
                }
```

nitz:
That looks eerily similar to what I just pushed in #120 🙂, if you wanted to check that out!

CartBlanche:
Thank again for your PRs, Closing this as we went with your WMI solution.

PR #112, opened 2021-11-13
CLI consolidatation

adrianstevens:


Meadow.CLI.Core/Devices/MeadowDeviceHelper.cs
```diff
@@ -524,13 +522,7 @@ await Task.Delay(1000, cancellationToken)
...
         private void Dispose(bool disposing)
         {
-            lock (Logger)
-            {
-                if (disposing)
-                {
-                    _meadowDevice?.Dispose();
-                }
-            }
+            _meadowDevice?.Dispose();
```
CartBlanche, line 525: This changes doesn't seem CLI Consolidation related. Are we sure that removing the lock won't cause race conditions?

PR #114, opened 2021-12-01
Implement minimum-codesize linker via the Mono IL Linker.

alexischr:
Deployment assemblies go through an extra processing step now, to remove unused code and save resources.


Meadow.CLI/Meadow.CLI.csproj
```diff
@@ -36,6 +36,11 @@
     <EmbeddedResource Remove="Internals\**" />
     <None Remove="Internals\**" />
   </ItemGroup>
+  <ItemGroup>
+    <Content Include="resources/monolinker.exe" CopyToOutputDirectory="Always" />
+    <Content Include="resources/meadow_link.xml" CopyToOutputDirectory="Always" />
+    <Content Include="resources/Mono.Cecil.dll" CopyToOutputDirectory="Always" />
```
CartBlanche, line 42: @alexischr Do we want a specific version of `Mono.Cecil` here, or would letting the nuget pull it in be ok? I'm thinking say down the line, do we want to update the required `Mono.Cecil` via a nuget update or manually?

alexischr: We can probably switch to the nuget in the future, but since we use an older Mono I thought it safer to hardlink to the Mono.Cecil generated by us, than try to find the right Mono.Cecil for our Mono 6.8 fork. On the other hand Mono.Cecil is probably super stable, so worth looking into

CartBlanche: Sure.

Meadow.CLI.Core/DeviceManagement/AssemblyManager.cs
```diff
@@ -15,6 +17,58 @@ public static class AssemblyManager
...
+            }
+
+            var postlink_dir = Path.Combine (path, "postlink_bin");
+
+            if (Directory.Exists(postlink_dir))
+                Directory.Delete(postlink_dir, recursive: true);
+            Directory.CreateDirectory (postlink_dir);
+
+            var base_path = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
+            var monolinker_path = Path.Combine (base_path, "resources/monolinker.exe");
```
adrianstevens, line 50: We need to remove the / here

var monolinker_path = Path.Combine (base_path, "resources", "monolinker.exe");
var descriptor_path = Path.Combine (base_path, "resources", "meadow_link.xml");

adrianstevens:
@alexischr I'm not able to run the included monolinker.exe on Windows - it appears it can't find the dll dependencies 

```Unhandled Exception: System.TypeLoadException: Could not load type 'System.Collections.Generic.Queue`1' from assembly 'mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'.```

adrianstevens:
and @CartBlanche - when this works, we'll need to include the monolinker files in the extension projects

CartBlanche:
I'm going to close this PR as you've implemented this in another PR.

PR #118, opened 2021-12-11
Add delete-all command

adrianstevens:


CartBlanche:
@adrianstevens How would I go about testing this, as the CLI side is VERY new to me? 
Also, Is it something we would expose through the extensions too?

thegreatco:
>Also, Is it something we would expose through the extensions too?

Anything we want exposed to anything other than the CLI needs to be (at a minimum) implemented under `MeadowDeviceHelper`, if we want it to be available via direct access to a Meadow, `IMeadowDevice`. 

I would argue that `IMeadowDevice` should probably be kept to the bare minimum of methods, basically only those commands implemented directly by `HCOM`. If we take this stance, we do have some cleaning up to do to follow this rule, for example `IsFileOnDevice` isn't an `HCOM` message, it is a higher level operation utilizing `GetFilesAndCrcsAsync`.

`MeadowDeviceHelper` is a wrapper around `IMeadowDevice` which should implement the higher level operations that may require multiple `HCOM` operations to complete. It also handles reconnecting when a `Meadow` restarts and finding the "right" `Meadow` when multiple devices are connected to the system.

@adrianstevens while this looks fine to me, I think we should consider moving this to `MeadowDeviceHelper` to make it shared code with the IDE plugins.

PR #120, opened 2021-12-14
Replace `wmic` tooling for `GetSerialPortsWindows()` with `System.Management` equivalent 

nitz:
As per discussion in #111, this is `GetSerialPortsForWindows()` using instead the .NET System.Management APIs. This introduces a dependency on the (Microsoft owned/provided) package `System.Management`, and implicitly `System.CodeDom`.

However, it is substantially faster than the invocation of an external executable and result parsing demonstrated in #111. It still provides the benefit of filtering the available ports to specifically find USB devices with Wilderness Labs' VID, which significantly reduces potential friction for any users with more than a single COM port (USB or otherwise.)


CartBlanche:
@nitz I'm liking this better than the PS solution. I have to try and get a new, more stable extension out, so will play with this after that. Probably Thursday or Friday now.

nitz:
Excellent!

Just as a point of reference, this way runs *so* much better on my machine compared to the other way. I have a relatively complex powershell profile that causes powershell to be slow to launch otherwise, so being able to cut that out saves an incredible amount of overhead.

nitz:
Gah I just noticed my auto-formatter messed with a bunch of code that was outside of the scope of what I was editing. Gonna force push an amended commit that undoes that.

CartBlanche:
@nitz Just to keep you in the loop: We've been busy preparing the next release, so I've not forgotten about this, so will be revisiting this shortly.

CartBlanche:
Closing as changes merged in other PR.

PR #135, opened 2022-03-12
Update erase flash version check logic

adrianstevens:
Fix for erase flash logic 

Compare to the mono version instead of the OS version since the OS version will reflect whatever has just been written to Meadow (i.e. we want the previous version not the current version)

### DIFF
--- a/Meadow.CLI/Commands/DeviceManagement/FlashOsCommand.cs
+++ b/Meadow.CLI/Commands/DeviceManagement/FlashOsCommand.cs
@@ -89,11 +89,13 @@ namespace Meadow.CLI.Commands.DeviceManagement
 
             Meadow = new MeadowDeviceHelper(meadow, Logger);
 
-            // Get Current Device Version
-            var currentOsVersion = new Version(Meadow.DeviceInfo?.MeadowOsVersion.Split(' ')[0]);
+            // Get Previous OS Version
+            // We just flashed the OS so it will show the current version 
+            // But the runtime hasn't been updated yet so should match the previous OS version
+            var previousOsVersion = new Version(Meadow.DeviceInfo?.MonoVersion.Split(' ')[0]);
 
             // If less that B6.1 flash
-            if (currentOsVersion.CompareTo(new Version(MINIMUM_OS_VERSION)) < 0) {
+            if (previousOsVersion.CompareTo(new Version(MINIMUM_OS_VERSION)) < 0) {
                 // Ask User 1st before wiping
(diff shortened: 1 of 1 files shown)


Meadow.CLI/Commands/DeviceManagement/FlashOsCommand.cs
```diff
@@ -89,11 +89,13 @@ public override async ValueTask ExecuteAsync(IConsole console)
...
-            // Get Current Device Version
-            var currentOsVersion = new Version(Meadow.DeviceInfo?.MeadowOsVersion.Split(' ')[0]);
+            // Get Previous OS Version
+            // We just flashed the OS so it will show the current version 
+            // But the runtime hasn't been updated yet so should match the previous OS version
+            var previousOsVersion = new Version(Meadow.DeviceInfo?.MonoVersion.Split(' ')[0]);
 
             // If less that B6.1 flash
-            if (currentOsVersion.CompareTo(new Version(MINIMUM_OS_VERSION)) < 0) {
+            if (previousOsVersion.CompareTo(new Version(MINIMUM_OS_VERSION)) < 0) {
```
CartBlanche, line 98: @adrianstevens and the `MINIMUM_OS_VERSION` is still valid for `MonoVersion`? 
Is that kept in sync with `MeadowOsVersion`?

CartBlanche (changes requested):

Meadow.CLI/Commands/DeviceManagement/FlashOsCommand.cs
adrianstevens: From Slack - the OS version and the runtime (mono) version are stored and retrieved separately but to date we've always released them together. Or said another way - the OS and mono versions should always match.





[11:30](https://wildernesslabs.slack.com/archives/CPWT8EZEK/p1647196213660139)
In the current CLI where we added the logic we're checking the OS version - but we're checking it after we've updated the OS so it will always get the version of the new OS, not the version we're replacing.
[11:32](https://wildernesslabs.slack.com/archives/CPWT8EZEK/p1647196351406069)
So it is possible that someone has mismatched mono and OS versions so checking the mono version isn't a perfect solution but it's highly unlikely that someone has a mono version installed that's greater than the OS and I can't think of any reasonable way to get into that stake.

CartBlanche: Ok cool. I just wanted to make sure.

PR #147, opened 2022-04-19
Remove logic to automatically use local runtime for flash os

adrianstevens:
This standardizes and simplifies - and this logic has been confusing for a couple of reasons:

1. we don't have comparable logic for the OS binary so it was only automatically using the runtime binary
2. we have explicit command line arguments to use local binaries for flash os


CartBlanche (approved):
LGTM

PR #148, opened 2022-04-19
Add ability to specify a downloaded OS version for flash os

adrianstevens:
e.g. `meadow flash os -v 0.6.3.1`

### DIFF
--- a/Meadow.CLI.Core/Devices/IMeadowDevice.cs
+++ b/Meadow.CLI.Core/Devices/IMeadowDevice.cs
@@ -22,6 +22,7 @@ namespace Meadow.CLI.Core.Devices
         public Task FormatFileSystemAsync(uint partition = 0, CancellationToken cancellationToken = default);
         public Task RenewFileSystemAsync(CancellationToken cancellationToken = default);
         public Task UpdateMonoRuntimeAsync(string? fileName, uint partition = 0, CancellationToken cancellationToken = default);
+        public Task UpdateMonoRuntimeAsync(string? fileName, string? osVersion, uint partition = 0, CancellationToken cancellationToken = default);
         public Task WriteFileToEspFlashAsync(string fileName, uint partition = 0, string? mcuDestAddress = null, CancellationToken cancellationToken = default);
         public Task FlashEspAsync(string? sourcePath, CancellationToken cancellationToken = default);
         public Task<MeadowDeviceInfo> GetDeviceInfoAsync(TimeSpan timeout, CancellationToken cancellationToken = default);
--- a/Meadow.CLI.Core/Devices/IMeadowDevice.cs
+++ b/Meadow.CLI.Core/Devices/IMeadowDevice.cs
@@ -24,7 +24,11 @@ namespace Meadow.CLI.Core.Devices
         public Task UpdateMonoRuntimeAsync(string? fileName, uint partition = 0, CancellationToken cancellationToken = default);
         public Task UpdateMonoRuntimeAsync(string? fileName, string? osVersion, uint partition = 0, CancellationToken cancellationToken = default);
         public Task WriteFileToEspFlashAsync(string fileName, uint partition = 0, string? mcuDestAddress = null, CancellationToken cancellationToken = default);
+
+        public Task FlashEspAsync(CancellationToken cancellationToken = default);
         public Task FlashEspAsync(string? sourcePath, CancellationToken cancellationToken = default);
+        public Task FlashEspVersionAsync(string osVersion, CancellationToken cancellationToken = default);
+
         public Task<MeadowDeviceInfo> GetDeviceInfoAsync(TimeSpan timeout, CancellationToken cancellationToken = default);
         public Task<string?> GetDeviceNameAsync(TimeSpan timeout, CancellationToken cancellationToken = default);
         public Task<bool> GetMonoRunStateAsync(CancellationToken cancellationToken = default);
@@ -55,7 +59,6 @@ namespace Meadow.CLI.Core.Devices
         public Task<string?> GetDeviceMacAddressAsync(CancellationToken cancellationToken = default);
         public Task<bool> InitializeAsync(CancellationToken cancellationToken);
         public bool IsDeviceInitialized();
-        public Task FlashEspAsync(CancellationToken cancellationToken = default);
         public Task<bool> IsFileOnDevice(string filename, CancellationToken cancellationToken = default);
     }
 }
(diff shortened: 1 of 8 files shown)


Meadow.CLI.Core/Devices/IMeadowDevice.cs
```diff
@@ -22,8 +22,13 @@ public interface IMeadowDevice : IDisposable
         public Task FormatFileSystemAsync(uint partition = 0, CancellationToken cancellationToken = default);
         public Task RenewFileSystemAsync(CancellationToken cancellationToken = default);
         public Task UpdateMonoRuntimeAsync(string? fileName, uint partition = 0, CancellationToken cancellationToken = default);
+        public Task UpdateMonoRuntimeAsync(string? fileName, string? osVersion, uint partition = 0, CancellationToken cancellationToken = default);
         public Task WriteFileToEspFlashAsync(string fileName, uint partition = 0, string? mcuDestAddress = null, CancellationToken cancellationToken = default);
+
+        public Task FlashEspAsync(CancellationToken cancellationToken = default);
         public Task FlashEspAsync(string? sourcePath, CancellationToken cancellationToken = default);
+        public Task FlashEspVersionAsync(string osVersion, CancellationToken cancellationToken = default);
```
alexischr, line 30: Could you explain why `UpdateMonoRuntimeAsync` got a method overrride while `FlashEspAsync` gets a variant?

alexischr:
I think there is some confusing usage of `osVersion` as sometimes something that is null or something that is an empty string. Just confirming that there is a reason why "version not defined" checks aren't merged into either null or ""

Meadow.CLI.Core/Devices/IMeadowDevice.cs
adrianstevens: it's because we don't currently have the ability to specify a filename or location for the ESP binaries - which we could do - but it'll require a fair bit of rework

adrianstevens:
@alexischr I updated the string params to nullable - thx!

Meadow.CLI.Core/Devices/MeadowDeviceHelper.cs
```diff
@@ -367,7 +367,7 @@ await Task.Delay(1000, cancellationToken)
             _meadowDevice = meadow ?? throw new Exception($"Meadow not found. Serial Number {serialNumber}");
         }
 
-        public async Task FlashOsAsync(string? runtimePath = null, bool skipRuntime = false, bool skipEsp = false, CancellationToken cancellationToken = default)
+        public async Task FlashOsAsync(string? runtimePath = null, string? osVersion = null, bool skipRuntime = false, bool skipEsp = false, CancellationToken cancellationToken = default)
```
CartBlanche, line 370: Just wondering why add the parameter here rather than at the end, so that existing API calls, with default params would just work?

Meadow.CLI.Core/Devices/MeadowDeviceHelper.cs
```diff
@@ -383,7 +383,7 @@ await MonoDisableAsync(cancellationToken: cancellationToken)
                                  "Meadow was expected to have Mono Disabled");
 
                     await _meadowDevice.UpdateMonoRuntimeAsync(
-                        runtimePath,
+                        runtimePath, osVersion,
```
CartBlanche, line 386: Nit: For consistency, maybe put the `osVersion` on it's own line?

Meadow.CLI.Core/Devices/IMeadowDevice.cs
```diff
@@ -22,8 +22,13 @@ public interface IMeadowDevice : IDisposable
         public Task FormatFileSystemAsync(uint partition = 0, CancellationToken cancellationToken = default);
         public Task RenewFileSystemAsync(CancellationToken cancellationToken = default);
         public Task UpdateMonoRuntimeAsync(string? fileName, uint partition = 0, CancellationToken cancellationToken = default);
-        public Task WriteFileToEspFlashAsync(string fileName, uint partition = 0, string? mcuDestAddress = null, CancellationToken cancellationToken = default);
+        public Task UpdateMonoRuntimeAsync(string? fileName, string? osVersion, uint partition = 0, CancellationToken cancellationToken = default);
+        public Task WriteFileToEspFlashAsync(string? fileName, uint partition = 0, string? mcuDestAddress = null, CancellationToken cancellationToken = default);
+
+        public Task FlashEspAsync(CancellationToken cancellationToken = default);
         public Task FlashEspAsync(string? sourcePath, CancellationToken cancellationToken = default);
+        public Task FlashEspVersionAsync(string? osVersion, CancellationToken cancellationToken = default);
```
CartBlanche, line 30: Rather than have a separate, Version specific API, could we not modify  FlashEspAsync to 
`public Task FlashEspAsync(CancellationToken cancellationToken = default, string? osVersion = null);`

I'm torn between having specific APIs and using existing APIs with default parameters. Open to chat about pros anc ons of each.

Meadow.CLI/Commands/DeviceManagement/FlashOsCommand.cs
```diff
@@ -49,7 +52,7 @@ public override async ValueTask ExecuteAsync(IConsole console)
 
             if (!SkipDfu)
             {
-                serialNumber = await MeadowDeviceHelper.DfuFlashAsync(SerialPortName, OsFile, Logger, cancellationToken).ConfigureAwait(false);
+                serialNumber = await MeadowDeviceHelper.DfuFlashAsync(SerialPortName, OsFile, Logger, cancellationToken, OSVersion).ConfigureAwait(false);
```
CartBlanche, line 55: With `DfuFlashAsync` we've added the `OSVersion` parameter to the end but with `FlashOsAsync` (below) it's the 2nd parameter.

Thoughts on if we should standardise where it goes?

Meadow.CLI.Core/Devices/MeadowDeviceHelper.cs
adrianstevens: fixed :)

Meadow.CLI.Core/Devices/IMeadowDevice.cs
adrianstevens: The problem is the default param for the cancelationToken .... but agreed, I don't love needing a new API

Meadow.CLI/Commands/DeviceManagement/FlashOsCommand.cs
adrianstevens: I think we should standardize - I'll fix this

Meadow.CLI.Core/Devices/IMeadowDevice.cs
adrianstevens: Fixed!

Meadow.CLI.Core/Devices/MeadowDeviceHelper.cs
CartBlanche: If we are standardising, shouldn't the `osVersion` be at the end on this one too?

Meadow.CLI/Commands/DeviceManagement/FlashOsCommand.cs
adrianstevens: fixed

Meadow.CLI.Core/Devices/MeadowDeviceHelper.cs
adrianstevens: fixed

CartBlanche (approved):
LGTM

PR #152, opened 2022-04-23
Adrian fix dfu fail

adrianstevens:


CartBlanche:
@adrianstevens since the other PR was merged, looks like this may need a rebase.

PR #156, opened 2022-05-20
Use version for esp path with flashing

adrianstevens:


Meadow.CLI.Core/Devices/MeadowLocalDeviceFileManager.cs
```diff
@@ -382,7 +382,7 @@ await SendTheEntireFile(command, lastFile, cancellationToken)
                 sourcePath ??= DownloadManager.FirmwareDownloadsFilePath;
             }
             else {
-                sourcePath ??= DownloadManager.FirmwarePathForVersion(osVersion);
+                sourcePath = DownloadManager.FirmwarePathForVersion(osVersion);
```
CartBlanche, line 385: Just to be clear, we don't care if `sourcePath` is not null, we always want it to use the passed in `*.FirmwarePath` etc?

adrianstevens: Yeah that property name is somewhat misleading .... its the default firmware path but we want to override it when specifying osVersion

CartBlanche: Cool!

PR #178, opened 2022-07-28
Use Non-blocking FileStream.ReadAsync(), rather than File.ReadAllBytes()

CartBlanche:


Meadow.CLI.Core/Devices/MeadowLocalDeviceFileManager.cs
```diff
@@ -113,7 +113,14 @@ await SendCommandAsync(command, cancellationToken)
 
             // If ESP32 file we must also send the MD5 has of the file
             using var md5 = MD5.Create();
-            var fileBytes = File.ReadAllBytes(sourceFileName);
+
+            byte[] fileBytes;
+            using (FileStream stream = File.Open(sourceFileName, FileMode.Open))
+            {
+                fileBytes = new byte[stream.Length];
+                await stream.ReadAsync(fileBytes, 0, (int)stream.Length)
```
thegreatco, line 121: Since you're not checking the result of `ReadAsync` would could, in theory, [get an incomplete read](https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.readasync?view=netstandard-2.1#system-io-stream-readasync(system-byte()-system-int32-system-int32)).
>A task that represents the asynchronous read operation. The value of the TResult parameter contains the total number of bytes read into the buffer. The result value can be less than the number of bytes requested if the number of bytes currently available is less than the requested number, or it can be 0 (zero) if the end of the stream has been reached.

One way around this would be to use a `MemoryStream` and `CopyToAsync`, something like:
```
using var md5 = MD5.Create();

byte[] fileBytesFromPool = ArrayPool<byte>.Shared.Rent((int)fi.Length);
try
{
    using (FileStream stream = File.Open(sourceFileName, FileMode.Open))
    using (MemoryStream ms = new MemoryStream(fileBytesFromPool))
    {
        await stream.CopyToAsync(ms)
                    .ConfigureAwait(false);
    }

    var fileBytes = fileBytesFromPool[..(int)fi.Length];
    var hash = md5.ComputeHash(fileBytes);
    string md5Hash = BitConverter.ToString(hash)
                                 .Replace("-", "")
                                 .ToLowerInvariant();

    var fileCrc32 = CrcTools.Crc32part(fileBytes, fileBytes.Length, 0);

    var command =
        await new FileCommandBuilder(
                  HcomMeadowRequestType.HCOM_MDOW_REQUEST_START_FILE_TRANSFER)
              .WithSourceFileName(sourceFileName)
              .WithDestinationFileName(destinationFileName)
              .WithTimeout(timeout)
              .WithPartition(0)
              .BuildAsync();

    var sw = Stopwatch.StartNew();
    await SendTheEntireFile(command, true, cancellationToken)
        .ConfigureAwait(false);

    sw.Stop();
    return new FileTransferResult(sw.ElapsedMilliseconds, fileBytes.Length, fileCrc32);
}
finally
{
    ArrayPool<byte>.Shared.Return(fileBytesFromPool);
}
```

But this requires us to use C# 8 to get access to `System.Range`/`System.Index`....

So I guess we can keep it simple stupid and deal with the minor allocation in `new byte[(int)fi.Length]`
```
byte[] fileBytes = new byte[(int)fi.Length];
using FileStream stream = File.Open(sourceFileName, FileMode.Open);
using MemoryStream ms = new MemoryStream(fileBytes);

await stream.CopyToAsync(ms)
            .ConfigureAwait(false);
```

I generally prefer we use `var` whenever possible, so `FileStream stream` -> `var fs` and `MemoryStream ms` -> `var ms` or similar.

thegreatco (changes requested):
I think in general we should start looking to upgrade to C# 8 and something newer than .netstandard2.0 if we can.

Meadow.CLI.Core/Devices/MeadowLocalDeviceFileManager.cs
```diff
@@ -113,7 +113,14 @@ await SendCommandAsync(command, cancellationToken)
 
             // If ESP32 file we must also send the MD5 has of the file
             using var md5 = MD5.Create();
-            var fileBytes = File.ReadAllBytes(sourceFileName);
```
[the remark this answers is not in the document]
CartBlanche: @thegreatco I had hoped to use that initially, but as you point out it's in 2.1 and I didn't think we could risk switching to 2.1 so close to v1.x, hence why I went with `ReadAsync`

Meadow.CLI.Core/Devices/MeadowLocalDeviceFileManager.cs
```diff
@@ -115,12 +115,17 @@ await SendCommandAsync(command, cancellationToken)
...
+            using (var stream = File.Open(sourceFileName, FileMode.Open))
             {
-                fileBytes = new byte[stream.Length];
-                await stream.ReadAsync(fileBytes, 0, (int)stream.Length)
+                var streamLength = (int)stream.Length;
+                fileBytes = new byte[streamLength];
+                var bytesRead = await stream.ReadAsync(fileBytes, 0, streamLength, cancellationToken)
                     .ConfigureAwait(false);
+                if (bytesRead != streamLength) {
+                    throw new InvalidDataException($"Read bytes: {bytesRead} from {sourceFileName} does not match stream Length: {streamLength}!");
```
thegreatco, line 125: This is likely to just cause random failures. We really should just loop until we've read it all, or copy to a memory stream.

thegreatco (dismissed):

Meadow.CLI.Core/Devices/MeadowLocalDeviceFileManager.cs
CartBlanche: @thegreatco We are getting the streamLength from the File.Open itself (which would think is correct as it's coming from the file system itself, presumably, which returns a FileStream. FileStream only has a `CanRead` property or Position property, AFAICT. Are you suggesting looping on either of those until it's false or the position matches the streamLength? There's no `DataAvailable` property, like we have with NetworkStreams.

thegreatco: I think you can just do something like:
```
using var stream = File.Open(sourceFileName, FileMode.Open);

var streamLength = (int)stream.Length;
var fileBytes = new byte[streamLength];
var bytesRead = 0;
while (stream.Position < stream.Length)
{
    bytesRead += await stream.ReadAsync(fileBytes, bytesRead, streamLength, cancellationToken)
                             .ConfigureAwait(false);
}

if (bytesRead != streamLength) {
    throw new InvalidDataException($"Read bytes: {bytesRead} from {sourceFileName} does not match stream Length: {streamLength}!");
}
```

CartBlanche: @thegreatco Implemented as you suggested. On Mac it works great. I've not had a chance to test it on Windows or in VSCode yet.

adrianstevens (approved):

PR #182, opened 2022-08-27
bring back the progress bar (for file push anyway)

ctacke:


CartBlanche:
@ctacke I had a quick play with it and unfortunately because we are using ILogger.* for our output, the Console.Write() isn't currently going to the right place, so nothing it actually displayed. Not sure if we need a custom Ilogger so that Log.Information or similar can accept the Write, rather than a WriteLine.

PR #186, opened 2022-09-03
Add logic to deploy non-linked mscorlib and system dlls + minor cleanup

adrianstevens:


Meadow.CLI.Core/Devices/MeadowLocalDeviceFileManager.cs
```diff
@@ -535,93 +541,115 @@ async Task DeleteTemporaryFiles(CancellationToken cancellationToken = default)
...
-                var dependencies = AssemblyManager.GetDependencies (fi.Name, fi.DirectoryName, osVersion)
-                    .Where (x => x.Contains ("App.") == false).ToList ();
+                var dependencies = AssemblyManager.GetDependencies(fi.Name, fi.DirectoryName, osVersion)
+                    .Where(x => x.Contains("App.") == false)
+                    .ToList();
 
-                var trimmed_dependencies = await AssemblyManager.TrimDependencies (fi.Name, fi.DirectoryName, dependencies, includePdbs: includePdbs);
+                var trimmedDependencies = (await AssemblyManager.TrimDependencies(fi.Name, fi.DirectoryName, dependencies, includePdbs: includePdbs))
+                    .Where(x => x.Contains("App.") == false)
+                    .Where(x => dllLinkIngoreList.Any(f => x.Contains(f)) == false)
```
CartBlanche, line 625: @adrianstevens those files will definitely be ignored, but are they being added to the  deploy list elsewhere?
I'm looking at this on my phone, so maybe I've missed it?

CartBlanche, line 625: @adrianstevens I think these ignore where clauses need to go on the `dependencies` list above. This will avoid the trimmer even processing them, thus saving a some cycles.

Meadow.CLI.Core/Devices/MeadowLocalDeviceFileManager.cs
```diff
@@ -535,93 +541,115 @@ async Task DeleteTemporaryFiles(CancellationToken cancellationToken = default)
...
+                    .Where(x => pdbLinkIngoreList.Any(f => x.Contains(f)) == false)
```
CartBlanche, line 626: @adrianstevens I don't think pdbs are picked up at this stage (i could be wrong), but added in later once the `files` list has been created.

PR #202, opened 2022-10-04
Add AnyCPU into the mix for testing.

CartBlanche:


adrianstevens (approved):

Meadow.CLI/Meadow.CLI.csproj
```diff
@@ -2,16 +2,16 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-	<TargetFrameworks>net5.0;net6.0</TargetFrameworks>
+	<TargetFrameworks>net5.0;</TargetFrameworks>
```
adrianstevens, line 5: In the future we should move this to NET 6.0 since 5.0 is no longer supported. But I fully support getting it working first :)

CartBlanche: Yes, that's my thinking right now

PR #219, opened 2022-11-20
Add path to ESP file name output

NevynUK:
Added the path to the output of the file name when transferring ESP files:
* Now consistent with OS and RT output
* Verification of actual file transferred is now possible.


CartBlanche (approved):
LGTM

PR #240, opened 2023-02-06
Fix Regression: Sometimes TrimmedDependencies will return null, when problems arise. …

CartBlanche:
…We need that so we can continue without trimming.


adrianstevens:
I want to take a careful look before we merge. This might negatively impact our hack to deploy a couple of specific dlls unlinked.

adrianstevens (approved):

Issue #229, opened 2022-12-08
Increase maximum packet size


NevynUK:
Increase the maximum size of the data packets that can be exchanged between the board and CLI.

Reason for request:
The device information packet is getting close to the maximum packet size.

NevynUK:
Forgot to mention that this change will need to be synchronised with OS similar OS change.

PR #245, opened 2023-02-13
Increase Max Allowable packet length

CartBlanche:
Fixes https://github.com/WildernessLabs/Meadow.CLI/issues/229


adrianstevens (approved):

CartBlanche:
I think we no longer need this as change was CPed into another PR.

PR #246, opened 2023-02-13
[OTA] add register device command

alexischr:


Meadow.CLI.Core/Devices/MeadowLocalDevice.cs
```diff
@@ -364,6 +364,22 @@ public async Task SetRtcTime(DateTimeOffset dateTime, CancellationToken cancella
...
+
+            var command =
+                new SimpleCommandBuilder(HcomMeadowRequestType.HCOM_MDOW_REQUEST_OTA_REGISTER_DEVICE)
+                    .WithResponseType(MeadowMessageType.Data)
+                    .Build();
+
+            var commandResponse =
+                await SendCommand(command, cancellationToken);
+
+            Console.WriteLine($"Device's public key: {commandResponse.Message}");
```
adrianstevens, line 379: @alexischr we can't use Console.WriteLine in the CLI - can you switch to the logger?

Meadow.CLI/Commands/Cloud/RegisterDeviceCommand.cs
```diff
@@ -0,0 +1,43 @@
...
+        public CloudRegisterCommand(DownloadManager downloadManager, ILoggerFactory loggerFactory, MeadowDeviceManager meadowDeviceManager)
+            : base(downloadManager, loggerFactory, meadowDeviceManager)
+        {
+        }
+
+        public async ValueTask ExecuteAsync(IConsole console)
+        {
+            var cancellationToken = console.RegisterCancellationHandler();
+
+            // var identityManager = new IdentityManager();
```
adrianstevens, line 27: Looks like this is still draft - so just a friendly note to clean up any unused code :)

Meadow.CLI.Core/Devices/MeadowLocalDevice.cs
alexischr: Yup, this is a work in progress with @lamebrain and I replacing the console output and disabled code with a push of the data to Meadow Cloud

alexischr:
Currently this new command returns right away, seemingly not sending the request. Investigating…

alexischr:
Above issue fixed (was copy/pasting seemingly bad code lacking an `override`)

Now the command times out waiting for the operation to conclude (RSA keygen takes a while on the STM32F7)

Meadow.CLI/Meadow.CLI.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
-	<TargetFrameworks>net5.0;net6.0</TargetFrameworks>
+	<TargetFrameworks>net5.0</TargetFrameworks>
```
adrianstevens, line 5: We'll need to discuss .NET versions, the last beta release is .NET 6.0 only and with .NET 5.0 being out of support we should probably move away from it.

Meadow.CLI.Core/Devices/MeadowLocalDevice.cs
lamebrain: meadow cloud portion added now

CartBlanche:
Looks like quite a few of the commits are already in `develop` . Could a rebase be missing?

PR #258, opened 2023-03-14
add package commands

lamebrain:
* refactored dependency injections for services and logger
* added appsettings.json
* package commands
  * create
  * upload
  * list
  * publish


Meadow.CLI/Properties/launchSettings.json
```diff
@@ -30,7 +30,7 @@
     },
     "FlashEsp": {
       "commandName": "Project",
-      "commandLineArgs": "flash esp"
+      "commandLineArgs": "flash esp -v 0.9.6.1"
```
CartBlanche, line 33: Do we want the version number hard coded here?

patridge (approved):
Not sure I know enough of the project to properly review this, but it looks good at the level I understand it.

Meadow.CLI/Properties/launchSettings.json
lamebrain: nope! thanks

Meadow.CLI.Core/Managers/PackageManager.cs
```diff
@@ -27,13 +27,13 @@ public string CreatePackage(string applicationPath, string osVersion)
...
             {
-                _logger.LogInformation($"\"{applicationPath}\" is not found. Please try again.");
+                throw new ArgumentException("Invalid applicationPath");
             }
 
             var osFilePath = Path.Combine(DownloadManager.FirmwareDownloadsFilePathRoot, osVersion);
             if (!Directory.Exists(osFilePath))
             {
-                _logger.LogInformation($"Version {osVersion} not found. Please download.");
+                throw new ArgumentException("osVersion not found. Please download.");
```
CartBlanche, line 36: In this thrown exception and the one above, why wouldn't we want the actual OSVerion or Application Path in the message?

lamebrain: we would want it :)

lamebrain:
> @lamebrain I've added a couple of questions, but I don't feel I'm familiar enough with the actual cloud requirements and functionality to comment on that.

Thanks! I've been trying to keep in mind that one day this will also be integrated into the IDEs, hence some of the service/manager refactorings. I've been locally testing a lot of the cloud functionality, so I feel ok about that as long as there isn't anything terribly wrong with the structure/design.

PR #260, opened 2023-03-18
Send mpak crc and file size

lamebrain:
* Add Crc32 Hasher class
* Update package service to add mpak crc and file size


Meadow.CLI.Core/Common/Hasher.cs
CartBlanche, line 1: Currently the Windows build is failing because this file hasn't been added to the Meadow.CLI.Core.VS2019.csproj. 
[Edit] See comment below about other CRC function/methods.

Meadow.CLI.Core/Common/Hasher.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using System.IO;
+using System.IO.Hashing;
+using System.Threading.Tasks;
+
+public class Hasher
+{
+    public async Task<string> CalculateCrc32(string filePath)
```
CartBlanche, line 8: Not sure if you are aware, but we have some CRC related functions/methods in the  `Meadow.CLI.Core.DeviceManagement.Tools` namespace. If there isn't anything you can already use there, maybe move this into that namespace so all the CRC helper functions are centralised?

CartBlanche (changes requested):

Meadow.CLI.Core/Common/Hasher.cs
lamebrain: Ah thanks, I was hoping something already existed and did a search, but came up empty. I'll find and use use the existing method as to avoid bringing in new dependencies.

Meadow.CLI.Core/Common/Hasher.cs
lamebrain: Yeah, I saw that and was saddened. Will refactor to the existing method. Thanks!

lamebrain:
@CartBlanche alright, I ended up moving the new function into the existing namespace. VS2019 also fixed.

CartBlanche (approved):
LGTM

PR #261, opened 2023-03-20
fixes for package commands

lamebrain:
* change AddDevice sig
* use `ProcessorId` instead of `SerialNumer` for device id provisioning
* move `cloud provision device` to `device provision`
* fix `package list` output


CartBlanche (approved):
LGTM. Just want to be clear I'm eye balling these changes and not had ability to test them. Hoping others, with more knowledge in the matter, will do that.

PR #275, opened 2023-04-04
Support previous HCOM protocol version

CartBlanche:
To test this
`meadow flash os -v 0.9.4.0`

then use normal commands and it should auto-downgrade to the previous protocol version, without throwing up the repeat error message.


patridge (approved):
A bit out of my expertise, but looks good to me.

CartBlanche:
@adrianstevens did you get a chance to try this?

adrianstevens (approved):

PR #277, opened 2023-04-11
Add Types to handle download failed message from Meadow.

CartBlanche:


Peter-Moody (approved):
Looks okay to me (considering I'm not using C# much these days).

Issue #278, opened 2023-04-12
Stdout from Meadow is output as ASCII instead of UTF-8


bcr:
My fancy Meadow app has UTF-8 output. This is turned into ASCII when monitoring with `meadow listen`. I know where to make the change, but I think it may affect other stuff. Works on my machine when using `meadow listen` but will probably explode if presented with non-UTF8 bytes with the high bit set.

https://github.com/WildernessLabs/Meadow.CLI/blob/develop/Meadow.CLI.Core/Internals/MeadowCommunication/ReceiveClasses/ReceiveSimpleText.cs#L31

CartBlanche:
@bcr what languages are you using?

bcr:
English. I was using the Degree Fahrenheit character \u2109.

CartBlanche:
@bcr I've create a PR here - https://github.com/WildernessLabs/Meadow.CLI/pull/279 
If you are feeling brave, this will create a Meadow.CLI  artefact here https://github.com/WildernessLabs/Meadow.CLI/actions/runs/4679452989 when finished, which you can install and test as follows, after you've downloaded it an extracted the zip file, grab the nuget from `Meadow.CLI/bin/Release` folder named WildernessLabs.Meadow.CLI.0.97.0-alpha.nupkg...

* dotnet tool uninstall --global WildernessLabs.Meadow.CLI
* dotnet tool install --global --add-source [your-download-dir] WildernessLabs.Meadow.CLI --version 0.97.0-alpha

if you find this alpha version unstable and want to go back to the latest stable version:

* dotnet tool uninstall --global WildernessLabs.Meadow.CLI
* dotnet tool install --global WildernessLabs.Meadow.CLI

bcr:
Yeah, that's basically what I did initially and it did what I expected. I submitted #280 which might be a more timid approach.

I just run it with `dotnet run --framework net6.0 -- listen` and it works -- do you prefer that I install it as a tool to test it?

CartBlanche:
@bcr if you have the cycles to install it as a tool, that would be great. Just to cover all bases.

CartBlanche:
Fixed in #279.

PR #280, opened 2023-04-12
Change StdOut parsing charset to UTF-8 from ASCII

bcr:
Proposed fix for #278. This is similar to #279 but only changes parsing for StdOut to use UTF-8.


CartBlanche:
@bcr thanks for sending this PR, greatly appreciated. I'll take a look and play with it later today.

CartBlanche:
@bcr thanks again for the more granular PR. For the time being we've merged the one line version which switches all output to UTF8 (for now) - We have new builds of all of the tooling available here for testing, if you have the cycles to give it a try -  https://github.com/WildernessLabs/Meadow.CLI/actions/runs/4736056630

So I'll close this PR for now, but should we find the "all in" version breaks things, we'll reconsider this PR.

PR #286, opened 2023-05-02
Add pre-release instructions

patridge:
Even if it's just for internal reference, I've added docs on how to install a downloaded pre-release version, such as those pulled from Actions builds.


CartBlanche (approved):
LGTM

PR #288, opened 2023-05-03
Revert "Revert "Changes to pass developer level to the device. (#259)…

CartBlanche:
…" (#265)"

This reverts commit bb79ffb0aabdd597396d24544a277ea39e5fa6fe.


NevynUK (approved):
Performed the following:

```bash
gh pr checkout 288
dotnet build
meadow set developer -d 0
```

The set developer command returned no tests which is expected.

I followed up with a breakpoint in `hcom_developer_tests_developer` and the `level` and `value` parameters looks correct for the command

```bash
meadow set developer -d 1002 -v 10
```

I think this is good to go.

PR #292, opened 2023-05-09
update package create

lamebrain:
* ability for app only
* ability for os only
* ability for app dir or file

### DIFF
--- a/Meadow.CLI.Core/Managers/PackageManager.cs
+++ b/Meadow.CLI.Core/Managers/PackageManager.cs
@@ -35,13 +35,13 @@ namespace Meadow.CLI.Core
                 else
                 {
                     var fi = new FileInfo(applicationPath);
-                    if(fi.Attributes.HasFlag(FileAttributes.Directory))
+                    if ((fi.Attributes & FileAttributes.Directory) == FileAttributes.Directory)
                     {
                         appFiles = Directory.GetFiles(applicationPath);
                     }
                     else
                     {
-                        appFiles = new List<string>{ fi.FullName }.ToArray();
+                        appFiles = new[] { fi.FullName };
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


Meadow.CLI.Core/Managers/PackageManager.cs
```diff
@@ -22,36 +23,67 @@ public PackageManager(ILoggerFactory loggerFactory)
...
+                else
+                {
+                    var fi = new FileInfo(applicationPath);
+                    if(fi.Attributes.HasFlag(FileAttributes.Directory))
+                    {
+                        appFiles = Directory.GetFiles(applicationPath);
+                    }
+                    else
+                    {
+                        appFiles = new List<string>{ fi.FullName }.ToArray();
```
CartBlanche, line 44: This will have a slight performance overhead because you are creating a new List, to add 1 element, then converting that list to an Array, when you can create the array directly, like :
`appFiles = new[] { fileInfo.FullName };`

Meadow.CLI.Core/Managers/PackageManager.cs
```diff
@@ -22,36 +23,67 @@ public PackageManager(ILoggerFactory loggerFactory)
...
-            {
-                throw new ArgumentException($"osVersion {osVersion} not found. Please download.");
+                if (!File.Exists(applicationPath) && !Directory.Exists(applicationPath))
+                {
+                    throw new ArgumentException($"Invalid applicationPath: {applicationPath}");
+                }
+                else
+                {
+                    var fi = new FileInfo(applicationPath);
+                    if(fi.Attributes.HasFlag(FileAttributes.Directory))
```
CartBlanche, line 38: I've not looked at how often CreatePackage is called, but something for the future, we can get a small perf improvement by using bitwise check, rather than doing the `HasFlag()` method call
        `if ((fi.Attributes & FileAttributes.Directory) == FileAttributes.Directory)`

Meadow.CLI.Core/Managers/PackageManager.cs
lamebrain: nice!

Meadow.CLI.Core/Managers/PackageManager.cs
lamebrain: nice!

CartBlanche (approved):

PR #300, opened 2023-05-29
Create package with / path separator always

Redth:
Windows `CreateEntryFromFile` will use the `\` windows path separator character when writing the zip file entry.  This might be fine and technically within spec, but when the update is extracted on the device it does not create the correct folder structure (so it seems `app\file.ext` is considered a file instead of a file within the `app` folder). See: https://github.com/dotnet/runtime/issues/41914

This just ensures on windows, the `/` separator replaces `\` instances when creating the package to avoid the issue.

### DIFF
--- a/Meadow.CLI.Core/Managers/PackageManager.cs
+++ b/Meadow.CLI.Core/Managers/PackageManager.cs
@@ -72,7 +72,7 @@ namespace Meadow.CLI.Core
                     {
                         foreach (var fPath in appFiles)
                         {
-                            archive.CreateEntryFromFile(fPath, Path.Combine("app", Path.GetFileName(fPath)));
+                            CreateEntry(archive, fPath, Path.Combine("app", Path.GetFileName(fPath)));
                         }
                     }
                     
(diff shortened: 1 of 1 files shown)


Meadow.CLI.Core/Managers/PackageManager.cs
```diff
@@ -85,5 +85,17 @@ public string CreatePackage(string applicationPath, string osVersion)
                 return string.Empty;
             }
         }
+
+        void CreateEntry(ZipArchive archive, string fromFile, string entryPath)
+        {
+            // Windows '\' Path separator character will be written to the zip which meadow os does not properly unpack
+            //  See: https://github.com/dotnet/runtime/issues/41914
+            if (OperatingSystem.IsWindows())
```
CartBlanche, line 93: @Redth sorry it's taken a while to look at this. So `IsWindows` a .NET 6 Property?

CartBlanche: I changed the check to use the runtimeServices and build is now green.

CartBlanche:
@Redth could you please review the change I made.

Redth:
@CartBlanche LGTM!

PR #310, opened 2023-06-04
ability to download os multiple times

lamebrain:


adrianstevens:
Let's hold on this since we have branching behavior depending on if the user specifies the version.

And we have a -f param to force a download, and this ignores that param.

adrianstevens (approved):

PR #313, opened 2023-06-12
Format VS4M release tag to new format.

CartBlanche:


CartBlanche:
Wait until tomorrow to pull in, just in case the extensions don't appear in the gallery.

PR #322, opened 2023-07-06
add support for app.build.yaml

ctacke:


adrianstevens (commented):
Looks reasonable to me but I haven't tested

CartBlanche (approved):
LGTM

CartBlanche:
@ctacke We could possibly use a stream reader to read the yaml file thus avoiding reading everything into a string holding variable, before passing it to the deserailizer. Something like:
`var opts = deserializer.DeserializeFromStream(File.OpenRead(buildOptionsFile));`

As John as tested your code, I didn't want to push another change in and start another test cycle, but something we might want to do down the line, perhaps.

PR #336, opened 2023-08-04
Reduce compilation warnings, so we can see when something is wrong.

CartBlanche:


jorgedevs:
@CartBlanche are you still working on this? If you're not, mark as Ready for Review and we can take a look.

CartBlanche:
@jorgedevs Yes I'm still working on it (2 other PRs have take priority over the last week). Also more testing required as it's quite an extensive change. Hopefully this week I'll change it to ready for review.

CartBlanche:
@jorgedevs I've decided to park this PR and instead work on zeroing the warnings in V2. 
It makes more sense to focus on that right now.

PR #340, opened 2023-08-11
update login redirect

lamebrain:
after `meadow cloud login`:
* redirect to meadowcloud
* call `api/users/me` to automatically create account in meadow cloud.

### DIFF
--- a/Meadow.CLI.Core/Identity/IdentityManager.cs
+++ b/Meadow.CLI.Core/Identity/IdentityManager.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks;
 using CredentialManagement;
 using IdentityModel.Client;
 using IdentityModel.OidcClient;
+using Microsoft.Extensions.Configuration;
 using Microsoft.IdentityModel.Logging;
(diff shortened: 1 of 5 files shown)


adunndevster2 (approved):

Meadow.CLI.Core/Identity/IdentityManager.cs
```diff
@@ -10,6 +10,7 @@
 using CredentialManagement;
 using IdentityModel.Client;
 using IdentityModel.OidcClient;
```
CartBlanche, line 12: @lamebrain I asked on [Slack](https://wildernesslabs.slack.com/archives/CPN84BWL9/p1691107733222179), and mentioned that this version of IdentityModel.OidcClient is no longer available in the nuget marketplace and if there is another replacement or is it worth forking the repo and strong signing it ourselves for inclusion with our own releases. Also avoids a compiler warning.

lamebrain: Thanks I created a ticket. Looks like there are other libs out there.

PR #358, opened 2023-09-28
Format device info out, to what we have now.

CartBlanche:


CartBlanche:
@ctacke AFAICT we don't support previous protocol versions in V2. So right now I'm doing this:
https://github.com/WildernessLabs/Meadow.CLI/pull/358/files#diff-f79cbc109b03e654e71c1fb3ad3719708f06103951cfa869c3c6cc67982efb0dR88
Is that something we will add?

ctacke:
I'm fine with not supporting previous versions.  This will require some rework because those fields don't necessarily exist in Linux but I can worry about that later.

ctacke (approved):

PR #364, opened 2023-09-30
Mono to Runtime

jorgedevs:


Source/v2/Meadow.Hcom/Connections/SerialConnection.cs
```diff
@@ -489,10 +489,10 @@ protected override void Dispose(bool disposing)
     private List<string> StdErr { get; } = new List<string>();
     private List<string> InfoMessages { get; } = new List<string>();
 
-    private const string RuntimeSucessfullyEnabledToken = "Meadow successfully started MONO";
```
ctacke, line 492: These can't be changed.  These are messages HCOM on the device sends to us and we use for parsing.

Source/v2/Meadow.Hcom/DeviceInfo.cs
```diff
@@ -71,19 +71,19 @@ public override string ToString()
...
-                    deviceInfo.AppendLine("    MAC Address - " );
+                    deviceInfo.AppendLine("    MAC Address - ");
                 }
                 deviceInfo.AppendLine($"{macAddresses}");
             }
 
             deviceInfo.AppendLine();
             deviceInfo.AppendLine($"Firmware Versions ");
             deviceInfo.AppendLine($"    OS: {OsVersion}");
-            deviceInfo.AppendLine($"    Mono: {RuntimeVersion}");
```
ctacke, line 86: this change is valid

ctacke (changes requested):

CartBlanche (approved):
LGTM

PR #385, opened 2023-10-21
proper handling of no connection/no route/use DFU

ctacke:


bryancostanich:
seems to be working for me now

bryancostanich (approved):

PR #389, opened 2023-10-25
Feature/local connection

ctacke:


CartBlanche:
@ctacke these changes look fairly straight forward. Are these changes to support Meadow.Windows/Linux?

PR #391, opened 2023-11-01
Make App Commands behave consistently...

CartBlanche:


Source/v2/Meadow.CLI/Commands/Helper/TaskExtensions.cs
```diff
@@ -0,0 +1,50 @@
+﻿using CliFx.Infrastructure;
+
+namespace Meadow.CLI
+{
+    public static class TaskExtensions
```
adrianstevens, line 5: I feel like this naming isn't quite right since it's not really a Task extension but a UI convenience 

Also not sure it should be an extension method but I understand the convenience

adrianstevens (approved):

Source/v2/Meadow.CLI/Commands/Helper/TaskExtensions.cs
CartBlanche: @adrianstevens I put it in TaskExtensions because it does essentially extend any `Task`, but happy to rename it to something more specific, like UITaskExtensions, for clarity.

CartBlanche: Done.

PR #401, opened 2023-11-17
Do a directory existing check before we check for internet.

CartBlanche:


Meadow.CLI.Core/Managers/DownloadManager.cs
```diff
@@ -171,19 +163,14 @@ public async Task DownloadOsBinaries(string? version = null, bool force = false,
...
             }
-            else
-            {
-                local_path = Path.Combine(FirmwareDownloadsFilePathRoot, version);
-            }
 
-            if (CreateFolder(local_path, force) == false)
+            local_path = Path.Combine(FirmwareDownloadsFilePathRoot, version);
+
+            if (!CreateFolder(local_path, force) && Directory.EnumerateFiles(local_path).Any())
```
adrianstevens, line 173: Yeah this is interesting ... at least validate there's one file. Part of me wonders if we just remove the "force" and always overwrite. I wonder if we're providing any value with the force flag? (something to chat about at least)

adrianstevens (approved):

PR #403, opened 2023-11-23
improve dependency filter for App.dll

ctacke:


Meadow.CLI.Core/Devices/MeadowLocalDevice.FileManager.cs
```diff
@@ -651,9 +651,8 @@ async Task AddFile(string file, bool includePdbs)
                 }
 
                 var dependencies = AssemblyManager.GetDependencies(fileName, directoryName, osVersion)
-                    .Where(x => x.Contains("App.") == false)
-                    // .Where(x => dllLinkIngoreList.Any(f => x.Contains(f)) == false)
-                    .ToList();
+                                    .Except(new string[] { "App.dll", "App.exe" })
```
stevenkuhn, line 654: Thoughts on having it be case-insensitive with `.Except(new string[] { "App.dll", "App.exe" }, StringComparer.OrdinalIgnoreCase)`?

adrianstevens (approved):

Meadow.CLI.Core/Devices/MeadowLocalDevice.FileManager.cs
adrianstevens: great call

adrianstevens: I don't remember if the OS itself requires it to be "App" but can't hurt regardless

PR #418, opened 2023-12-20
Linker refactor and v2 refresh

adrianstevens:


ctacke (approved):

CartBlanche:
@adrianstevens apparently your changes make use of "primary constructors" which is currently in preview. Keep in mind that the Windows extension is still net 4.7.2 (I'm looking into seeing if MS have pulled their finger out to allow newer .NET versions in extensions), so even though the CLI might build, the extension might not. I was recently bitten by a similar issue in another PR :/.

adrianstevens:
Absolutely - no need for the primary ctor, I was just playing with the feature

adrianstevens:
and fixed

PR #421, opened 2023-12-28
Fix Debug.WriteLine exception issue

duduita:
- https://github.com/WildernessLabs/Meadow_Issues/issues/3

Skipping the `System.Configuration` trimming fixes the `Debug.WriteLine()` exception. It increases the `.dll` size by ~60kB, which seems like an acceptable workaround side effect.


ctacke:
You can achieve this without code change by adding the assembly to `meadow_link.xml`:
https://github.com/WildernessLabs/Meadow.CLI/blob/develop/Meadow.CLI.Core/lib/meadow_link.xml

duduita:
Thank you for the hint @ctacke. Done!

alexischr (approved):

alexischr:
CLI errors seem unrelated - but is CLI CI broken?

PR #426, opened 2024-01-16
Fixes cleanup

adrianstevens:


CartBlanche (approved):
Untested, but changes look fairly straight forward. I'll let @ctacke do another pass on it, to have extra eyes on it.

PR #431, opened 2024-01-18
Warnings cleanup

adrianstevens:


Source/v2/Meadow.Cli/AppManager.cs
```diff
@@ -77,9 +77,9 @@ private static bool IsXmlDoc(string file)
             .Except(localFiles.Keys
                 .Select(f => Path.GetFileName(f)));
 
-        if (removeFiles.Count() == 0)
+        if (!removeFiles.Any())
```
CartBlanche, line 80: Really like this little optimisation! I keep forgetting Any is better than Count in this situation.

CartBlanche (approved):
Again untested, but changes LGTM. I'll let Chris do another pass to merge it in.

PR #434, opened 2024-01-23
Output cleanup, spinner, and updated app path logic

adrianstevens:


Source/v2/Meadow.Cli/Commands/Current/App/AppTrimCommand.cs
```diff
@@ -55,8 +53,13 @@ protected override async ValueTask ExecuteCommand()
         }
 
         // if no configuration was provided, find the most recently built
-        Logger?.LogInformation($"Trimming {file.FullName} (this may take a few seconds)...");
+        Logger?.LogInformation($"Trimming {file.FullName}");
+        Logger?.LogInformation("This may take a few seconds...");
+
+        var cts = new CancellationTokenSource();
+        ConsoleSpinner.Spin(Console, cancellationToken: cts.Token);
```
ctacke, line 60: Consider, perhaps, 
```
class ConsoleSPinner : IDisposable {}

using new ConsoleSpinner(Console);
spinner.Spin():
_packageManager.Trim()....
```
Not sure of any benefits or drawbacks over what you have, but it's an idea (that I had in some version of CLI somewhere, at some point)

ctacke (approved):

Source/v2/Meadow.Cli/Commands/Current/App/AppTrimCommand.cs
adrianstevens: Oh that's a great idea - i'll update

PR #439, opened 2024-01-30
[v2] Update meadow firmware download command to use Meadow Cloud

stevenkuhn:
This updates the **v2** version of `meadow firmware download` to retrieve the Meadow firmware from Meadow.Cloud. From the user's perspective the only change is that it **_will require a user to have a Meadow.Cloud account_** and for that user to login with `meadow cloud login`. The only additional command-line option is the `--host` argument:

```
USAGE
  meadow firmware download [options]

DESCRIPTION
  Download a firmware package

OPTIONS
  -f|--force        Default: "False".
  -v|--version
  --host            Optionally set a host (default is https://www.meadowcloud.co)
  -h|--help         Shows help text.
```

## Example

An example of this command would be:

```
C:\>meadow login
Logging into https://www.meadowcloud.co...
Signed in as user@example.org
Done

C:\>meadow firmware download
Retrieving firmware information from Meadow.Cloud...
Latest available version is '1.8.0.0'...
Downloading firmware package '1.8.0.0'...
Firmware package '1.8.0.0' downloaded
```

An example if the user is not logged in:
```
C:\>meadow firmware download
Retrieving firmware information from Meadow.Cloud...
You must be signed into Meadow.Cloud to execute this command. Run 'meadow cloud login' to do so.
```

### DIFF
--- a/Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
+++ b/Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
@@ -1,15 +1,26 @@
 ﻿using CliFx.Attributes;
+using CliFx.Exceptions;
+using Meadow.Cloud.Identity;
 using Meadow.Software;
 using Microsoft.Extensions.Logging;
+using System.Net.Http.Headers;
 
 namespace Meadow.CLI.Commands.DeviceManagement;
 
 [Command("firmware download", Description = "Download a firmware package")]
 public class FirmwareDownloadCommand : BaseFileCommand<FirmwareDownloadCommand>
 {
-    public FirmwareDownloadCommand(FileManager fileManager, ISettingsManager settingsManager, ILoggerFactory loggerFactory)
+    private IdentityManager IdentityManager { get; }
+
+    public FirmwareDownloadCommand(
+        FileManager fileManager,
+        IdentityManager identityManager, 
+        ISettingsManager settingsManager,
+        ILoggerFactory loggerFactory)
         : base(fileManager, settingsManager, loggerFactory)
-    { }
+    {
+        IdentityManager = identityManager;
+    }
 
     [CommandOption("force", 'f', IsRequired = false)]
     public bool Force { get; init; }
@@ -17,8 +28,24 @@ public class FirmwareDownloadCommand : BaseFileCommand<FirmwareDownloadCommand>
     [CommandOption("version", 'v', IsRequired = false)]
     public string? Version { get; set; }
 
+    [CommandOption("host", Description = "Optionally set a host (default is https://www.meadowcloud.co)", IsRequired = false)]
+    public string? Host { get; set; }
+
     protected override async ValueTask ExecuteCommand()
     {
+        Host ??= BaseCloudCommand<FirmwareDownloadCommand>.DefaultHost;
+
+        Logger?.LogInformation($"Retrieving firmware information from Meadow.Cloud{(Host != BaseCloudCommand<FirmwareDownloadCommand>.DefaultHost ? $" ({Host.ToLowerInvariant()})" : string.Empty)}...");
+
+        var token = await IdentityManager.GetAccessToken(CancellationToken);
+        if (string.IsNullOrWhiteSpace(token))
+        {
+            throw new CommandException("You must be signed into Meadow.Cloud to execute this command. Run 'meadow cloud login' to do so.");
+        }
+
+        FileManager.MeadowCloudClient.BaseAddress = new Uri(Host);
+        FileManager.MeadowCloudClient.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", token);
+
         await FileManager.Refresh();
 
         // for now we only support F7
--- a/Source/v2/Meadow.SoftwareManager/FileManager.cs
+++ b/Source/v2/Meadow.SoftwareManager/FileManager.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.IO;
+using System.Net.Http;
 using System.Threading.Tasks;
-
 
 namespace Meadow.Software;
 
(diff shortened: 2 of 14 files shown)


Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
```diff
@@ -1,24 +1,51 @@
...
     protected override async ValueTask ExecuteCommand()
     {
+        Host ??= BaseCloudCommand<FirmwareDownloadCommand>.DefaultHost;
+
+        Logger?.LogInformation($"Retrieving firmware information from Meadow.Cloud{(Host != BaseCloudCommand<FirmwareDownloadCommand>.DefaultHost ? $" ({Host.ToLowerInvariant()})" : string.Empty)}...");
+
+        var token = await IdentityManager.GetAccessToken(CancellationToken);
+        if (string.IsNullOrWhiteSpace(token))
+        {
+            throw new CommandException("You must be signed into Meadow.Cloud to execute this command. Run 'meadow cloud login' to do so.");
```
adrianstevens, line 43: Typically we've been logging an error and returning vs throwing an exception

stevenkuhn: I'm not opposed to that, but the documentation for CliFx mentions that commands do not directly return exit codes, but instead need to be done via `CommandException`:

https://github.com/Tyrrrz/CliFx?tab=readme-ov-file#reporting-errors

> Commands in CliFx do not directly return exit codes, but instead communicate execution errors via `CommandException`. This special exception type can be used to print an error message to the console, return a specific exit code, and also optionally show help text for the current command

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
```diff
@@ -1,24 +1,51 @@
...
     public string? Version { get; set; }
 
+    [CommandOption("host", Description = "Optionally set a host (default is https://www.meadowcloud.co)", IsRequired = false)]
+    public string? Host { get; set; }
+
     protected override async ValueTask ExecuteCommand()
     {
+        Host ??= BaseCloudCommand<FirmwareDownloadCommand>.DefaultHost;
+
+        Logger?.LogInformation($"Retrieving firmware information from Meadow.Cloud{(Host != BaseCloudCommand<FirmwareDownloadCommand>.DefaultHost ? $" ({Host.ToLowerInvariant()})" : string.Empty)}...");
```
adrianstevens, line 38: Should this only be logged if the host param is set?

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
stevenkuhn: Does throwing an exception or returning an error status code cause issues in non-typical CLI environments (like on a Build server)? I could see maybe a possibility that the exception isn't logged via the Meadow.CLI's logger.

adrianstevens: Probably how we should be doing it - I personally don't love the manual nature of "log error and return". We should file an issue to review and potentially update.

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
stevenkuhn: My intention with this is to always show `Retrieving firmware information from Meadow.Cloud...` when `meadow firmware download` is called.

If `--host URL_GOES_HERE` is specified, then it outputs that in the message as `Retrieving firmware information from Meadow.Cloud (URL GOES HERE)...` as an indicator of what Meadow.Cloud environment it is using.

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
stevenkuhn: Sounds good. For consistency sake, should I update this and other Cloud-related commands to go with log error and return for now?

adrianstevens: I personally vote for consistency

stevenkuhn: Should be fixed now, thanks!

stevenkuhn: I'll update the rest of the Meadow.Cloud-related commands in a separate PR.

Source/v2/Meadow.SoftwareManager/FileManager.cs
```diff
@@ -23,11 +23,14 @@ public class FileManager
 
     public FirmwareStore Firmware { get; }
 
-    public FileManager()
+    public HttpClient MeadowCloudClient { get; }
```
adrianstevens, line 26: Should FileManager own an instance of the Meadow Cloud http client? This feels out of place here.

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
```diff
@@ -1,24 +1,52 @@
...
+
+        var token = await IdentityManager.GetAccessToken(CancellationToken);
```
adrianstevens, line 40: I believe this needs to be null checked - what if we're not using the IdentityManager?

Source/v2/Meadow.SoftwareManager/FileManager.cs
stevenkuhn: One option: Instead of `new FileManager(HttpClient meadowCloudClient)` and a HttpClient property, there could be a method that allows the configuration of MeadowCloudClient in FileManager. Something like 

```csharp
FileManager.ConfigureMeadowCloudClient(c =>
{
  c.BaseAddress = new Uri(Host);
  c.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer", token);
});
```

If having a HttpClient instance it weird, I could wrap a custom class around HttpClient or create an custom interface that is passed to FileManager instead.

PR #444, opened 2024-02-02
Fix firmware typo in output

adrianstevens:


ctacke:
This will be part of a PR I'm working on

PR #447, opened 2024-02-06
[v2] [refactor] Update meadow firmware download command to use Meadow Cloud

stevenkuhn:
This updates the **v2** version of `meadow firmware download` to retrieve the Meadow firmware from Meadow.Cloud. From the user's perspective the only change is that it **_will require a user to have a Meadow.Cloud account_** and for that user to login with `meadow cloud login`. The only additional command-line option is the `--host` argument:

```
USAGE
  meadow firmware download [options]

DESCRIPTION
  Download a firmware package

OPTIONS
  -f|--force        Default: "False".
  -v|--version
  --host            Optionally set a host (default is https://www.meadowcloud.co)
  -h|--help         Shows help text.
```

## Example

An example of this command would be:

```
C:\>meadow login
Logging into https://www.meadowcloud.co...
Signed in as user@example.org
Done

C:\>meadow firmware download
Retrieving firmware information from Meadow.Cloud...
Latest available version is '1.8.0.0'...
Downloading firmware package '1.8.0.0'...
Firmware package '1.8.0.0' downloaded
```

An example if the user is not logged in:
```
C:\>meadow firmware download
Retrieving firmware information from Meadow.Cloud...
You must be signed into Meadow.Cloud to execute this command. Run 'meadow cloud login' to do so.
```

### Details

This introduces a new `IMeadowCloudClient` interface that provides the starting point when working with operations that involve Meadow.Cloud. The current properties and methods on this interface look like below.

```csharp
public interface IMeadowCloudClient
{
    IApiTokenClient ApiToken { get; }
    ICollectionClient Collection { get; }
    ICommandClient Command { get; }
    IDeviceClient Device { get; }
    IFirmwareClient Firmware { get; }
    IPackageClient Package { get; }
    IUserClient User { get; }

    Task<bool> Authenticate(string? host = default, CancellationToken cancellationToken = default);
}
```

Each individual client on `IMeadowCloudClient` represents a operations that are specific to that client. For this pull request, the only client implemented is `IFirmwareClient` which is described below:

```csharp
public interface IFirmwareClient
{
    Task<IEnumerable<GetFirmwareVersionsResponse>> GetVersions(string type, CancellationToken cancellationToken = default);
    Task<GetFirmwareVersionResponse?> GetVersion(string type, string version, CancellationToken cancellationToken = default);
    Task<HttpResponseMessage> GetDownloadResponse(string url, CancellationToken cancellationToken = default);
    Task<HttpResponseMessage> GetDownloadResponse(Uri url, CancellationToken cancellationToken = default);
}
```

The implementation for these clients abstracts out the specific details of the underlying `HttpClient` calls providing a cleaner approach for users of `IMeadowCloudClient`. For example, getting a specific firmware version from Meadow.Cloud in `F7FirmwareDownloadManager.GetReleaseMetadata()` is now this:

```csharp
public async Task<F7ReleaseMetadata?> GetReleaseMetadata(string? version = null, CancellationToken cancellationToken = default)
{
    version = !string.IsNullOrWhiteSpace(version) ? version : "latest";
    var response = await _meadowCloudClient.Firmware.GetVersion("Meadow_Beta", version, cancellationToken);

    if (response == null)
    {
        return null;
    }

    return new F7ReleaseMetadata()
    {
        Version = response.Version,
        MinCLIVersion = response.MinCLIVersion,
        DownloadURL = response.DownloadUrl,
        NetworkDownloadURL = response.NetworkDownloadUrl
    };
}
```

And the usage of the client in `FirmwareDownloadCommand` is reduced to:
```csharp
var isAuthenticated = await _meadowCloudClient.Authenticate(Host, CancellationToken);
if (!isAuthenticated)
{
    Logger?.LogError($"You must be signed into Meadow.Cloud to execute this command. Run 'meadow cloud login' to do so.");
    return;
}
```

### Instantiation and Usage

The creation of `IMeadowCloudClient` in the CLI is done through the DI container. Without a DI container, it would look something like this (not putting in `CancellationToken` for brevity):

```csharp
var httpClient = new HttpClient();            // create & manage HttpClient appropriately.
var identityManager = new IdentityManager();
var userAgent = "my.example.useragent";       // MeadowCloudUserAgent.Cli and
                                              // MeadowCloudUserAgent.Workbench
                                              // are also available.
IMeadowCloudClient client = new MeadowCloudClient(httpClient, identityManager, userAgent);

if (!await client.Authenticate())
{
    // TODO: do something if the user is not authenticated. In the CLI, the user
    // is required to do a `meadow cloud login`.
}

var versions = await client.Firmware.GetVersions("Meadow_Beta");
```


Source/v2/Meadow.Cloud.Client/ApiTokens/ApiTokenClient.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Meadow.Cloud.Client.ApiTokens;
+
+public interface IApiTokenClient
```
stevenkuhn, line 3: This and the other `I<Thing>Client` interfaces and classes are empty as a way to show what the new clients will look like.

Source/v2/Meadow.Cloud.Client/Identity/IdentityManager.cs
```diff
@@ -1,37 +1,29 @@
...
-        _logger = logger;
+        _logger = logger ?? NullLogger<IdentityManager>.Instance;
     }
 
     /// <summary>
     /// Kick off login
     /// </summary>
     /// <returns></returns>
-    public async Task<bool> Login(string host, CancellationToken? cancellationToken = default)
+    public async Task<bool> Login(string host, CancellationToken cancellationToken = default)
```
stevenkuhn, line 26: `CancellationToken? cancellationToken = default` and `CancellationToken cancellationToken = default` are effectively the same thing, but this saves on having to do `?? CancellationToken.None` everywhere.

Source/v2/Meadow.Cloud.Client/MeadowCloudClient.cs
```diff
@@ -0,0 +1,68 @@
...
+        _firmwareClient = new Lazy<FirmwareClient>(() => new FirmwareClient(httpClient));
+
+        _httpClient = httpClient;
+        _identityManager = identityManager;
+        _logger = logger ?? NullLogger<MeadowCloudClient>.Instance;
+
+        _httpClient.DefaultRequestHeaders.Add("User-Agent", userAgent);
+    }
+
+    public IApiTokenClient ApiToken => throw new NotImplementedException("This client is not implemented yet. Please use the 'ApiTokenService' instead.");
```
stevenkuhn, line 44: The plan would be to replace the service implementations with these new client implementations.

Source/v2/Meadow.Cloud.Client/MeadowCloudResponse.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Meadow.Cloud.Client;
+
+public class MeadowCloudResponse(HttpStatusCode statusCode, IReadOnlyDictionary<string, IEnumerable<string>> headers)
```
stevenkuhn, line 3: I was originally `MeadowCloudResponse<>` in my refactor, but decided to return the result directly for successful responses. Leaving these here for now, but I may delete if there is no use for them

Source/v2/Meadow.Cloud.Client/MeadowCloudUserAgent.cs
```diff
@@ -0,0 +1,25 @@
+﻿namespace Meadow.Cloud.Client;
+
+public class MeadowCloudUserAgent
```
stevenkuhn, line 3: This provides a way to have both strongly typed user agents and a way to specify a custom user agent.

Source/v2/Meadow.Cloud.Client/Services/CommandService.cs
```diff
@@ -32,7 +28,7 @@ public CommandService(IdentityManager identityManager) : base(identityManager)
         if (!response.IsSuccessStatusCode)
         {
             var message = await response.Content.ReadAsStringAsync();
-            throw new MeadowCloudException(message);
+            throw new MeadowCloudClassicException(message);
```
stevenkuhn, line 31: I need to rename this back to `MeadowCloudException`. There's a new `MeadowCloudException`, but it doesn't make sense to have two different exceptions here.

stevenkuhn: Done.

adrianstevens (approved):

PR #454, opened 2024-02-08
Project split

adrianstevens:


Source/v2/Meadow.Hcom/Connections/SerialConnection.ListenerProc.cs
```diff
@@ -238,7 +238,7 @@ private async Task ListenerProc()
                                             throw new Exception("Data received for unknown file");
                                         }
 
-                                        _readFileInfo.FileStream.Write(udp.FileData);
+                                        _readFileInfo.FileStream.Write(udp.FileData, 0, udp.FileData.Length);
```
CartBlanche, line 241: It's probably an ok change (untested by me), but wondering if this belongs in a separate PR?

ctacke (approved):

Source/v2/Meadow.Cli/Commands/Current/App/AppDeployCommand.cs
```diff
@@ -72,7 +72,7 @@ protected override async ValueTask ExecuteCommand()
             if (!file.Exists)
             {
                 // it's a directory - we need to determine the latest build (they might have a Debug and a Release config)
-                var candidates = PackageManager.GetAvailableBuiltConfigurations(path, "App.dll");
+                var candidates = Package.PackageManager.GetAvailableBuiltConfigurations(path, "App.dll");
```
stevenkuhn, line 75: The `Package.` prefix shouldn't be needed since you are including `Meadow.Package` in the using statement. You should be able to remove this and everywhere else that has the `Package.` prefix.

Source/v2/Meadow.Linker/Linker/ILLinker.cs
```diff
@@ -71,7 +74,9 @@ public ILLinker(ILogger? logger = null)
                     _logger?.Log(LogLevel.Debug, "StandardOutput Contains: " + stdOutReaderResult);
                 }
 
-                await process.WaitForExitAsync();
+                //await process.WaitForExitAsync();
+
+                await Task.Run(() => process.WaitForExit());
```
stevenkuhn, line 79: What's happening here? I'm guessing `process.WaitForExitAsync()` doesn't exist in `netstandard2.0`? Can the commented out code be removed?

Source/v2/Meadow.Package/PackageInfo.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Text.Json.Serialization;
+
+namespace Meadow.Package;
+
+public record PackageInfo
```
stevenkuhn, line 5: You don't _have_ to have this class. You could reference the one in Meadow.Cloud.Client. But I could see the case for decoupling it (even though they are exactly the same right now).

Source/v2/Meadow.Package/PackageManager.cs
```diff
@@ -50,7 +54,7 @@ private bool CleanApplication(string projectFilePath, string configuration = "Re
             if (dataLine.Data != null)
             {
                 Debug.WriteLine(dataLine.Data);
-                if (dataLine.Data.Contains("Build FAILED", StringComparison.InvariantCultureIgnoreCase))
+                if (dataLine.Data.ToLower().Contains("build failed"))
```
stevenkuhn, line 57: Why did you change this? The problem with just `ToLower()` is that it is culture dependent. If you are sticking with it, I'd suggest `.ToLower(CultureInfo.InvariantCulture)`. Or go with `dataLine.Data.Contains("build failed", StringComparison.OrdinalIgnoreCase)`.

Source/v2/Meadow.Hcom/Connections/SerialConnection.ListenerProc.cs
adrianstevens: This is required for the downgrade to .NET Standard 2.0 - otherwise completely agree

Source/v2/Meadow.Linker/Linker/ILLinker.cs
adrianstevens: exactly - I'll remove the old code

Source/v2/Meadow.Package/PackageInfo.cs
adrianstevens: ah apologies - I meant to delete from Meadow.Cloud.Client

Source/v2/Meadow.Package/PackageManager.cs
adrianstevens: the answer is - .NET Standard 2.0 :)

Source/v2/Meadow.Cli/Commands/Current/App/AppDeployCommand.cs
adrianstevens: yeah we have a PackageManager in both HCom and the Package project ...

adrianstevens: oooh but I can delete the one in HCom

adrianstevens: fixed

Source/v2/Meadow.Linker/Linker/ILLinker.cs
adrianstevens: fixed

Source/v2/Meadow.Package/PackageInfo.cs
adrianstevens: actually you're write - it's in the dependency chain - I deleted the copy in Meadow.Package

Source/v2/Meadow.Package/PackageManager.cs
stevenkuhn: Oh my! That makes sense -- I totally thought that would still be available in netstandard2.0.

Thoughts on `.ToLowerInvariant()`?

adrianstevens: Added CultureInfo.InvariantCulture - thx!

Source/v2/Meadow.Dfu/Meadow.Dfu.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+	<LangVersion>10.0</LangVersion>
+  </PropertyGroup>
```
stevenkuhn, line 6: You need `<Nullable>enable</Nullable>` here and in Meadow.Firmware and in Meadow.Package.

adrianstevens: great - adding

adrianstevens: done - added implicit usings while I was in there

stevenkuhn (approved):
💯

PR #458, opened 2024-02-09
sort firmware version output.  Fix bug where firmware command was not…

ctacke:
… seeting cloud URI

### DIFF
--- a/Source/v2/Meadow.Cli/Commands/Current/BaseCommand.cs
+++ b/Source/v2/Meadow.Cli/Commands/Current/BaseCommand.cs
@@ -32,13 +32,18 @@ public abstract class BaseCommand<T> : ICommand
         catch (Exception ex)
         {
             Logger?.LogError(ex.Message);
-            throw new CliFxException(ex.Message, -1);
+            throw new CommandException(
+                message: ex.Message,
+                exitCode: 1,
+                innerException: ex);
         }
 
         if (CancellationToken.IsCancellationRequested)
         {
             Logger?.LogInformation($"Cancelled");
-            throw new CliFxException("Cancelled", -2);
+            throw new CommandException(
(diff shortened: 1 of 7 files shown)


Source/v2/Meadow.Cli/Commands/Current/BaseCommand.cs
```diff
@@ -31,12 +32,13 @@ public async ValueTask ExecuteAsync(IConsole console)
         catch (Exception ex)
         {
             Logger?.LogError(ex.Message);
-            return;
+            throw new CliFxException(ex.Message, -1);
```
stevenkuhn, line 35: The CliFx documentation recommends/mentions throwing `CommandException` in their docs.

Source/v2/Meadow.Cloud.Client/MeadowCloudClient.cs
```diff
@@ -29,11 +16,16 @@ public class MeadowCloudClient : IMeadowCloudClient
     private readonly HttpClient _httpClient;
     private readonly IdentityManager _identityManager;
     private readonly ILogger _logger;
-    
+
     public MeadowCloudClient(HttpClient httpClient, IdentityManager identityManager, MeadowCloudUserAgent userAgent, ILogger<MeadowCloudClient>? logger = default)
     {
         _firmwareClient = new Lazy<FirmwareClient>(() => new FirmwareClient(httpClient));
 
+        if (httpClient.BaseAddress == null)
```
stevenkuhn, line 24: This makes sense for now, but I have to figure out a nice way to pass in a hostname here. It's currently being in set in the Authenticate method which doesn't isn't needed for anonymous calls.

adrianstevens (approved):

stevenkuhn (approved):

PR #460, opened 2024-02-10
Add support to set created mpak file name from CLI

doingnz:
There is an existing -n option which sets the property MpakPath. This value needs to be passed into the method creating the mpak. Default value to null which will can be checked to then automatically assign a file name using DateTime.


doingnz:
I added a comment regarding code that would force the generated file to have a ".mpak" extension. Not sure if you want to force the extension, or let the user set the extension when they specify a custom mpak file. Users will quickly realise they need to add .mpak to the name when they first run the create command and use -n and don't specify the extension.

ctacke (approved):

PR #465, opened 2024-02-12
Fix -f flag when downloading OS and always write version.txt when not explicitly setting a version

adrianstevens:


Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
```diff
@@ -77,7 +75,7 @@ protected override async ValueTask ExecuteCommand()
             return;
         }
 
-        if (collection[Version] != null)
+        if (collection[Version] != null && Force == false)
```
stevenkuhn, line 78: A small nit: Can this be `!Force`?

adrianstevens: I personally prefer the long version ... I can parse it faster, the ! tends to blend into the variable name

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
```diff
@@ -61,6 +58,7 @@ protected override async ValueTask ExecuteCommand()
             }
 
             Logger?.LogInformation($"Latest available version is '{latest}'...");
+            await collection.SetDefaultPackage(latest);
```
stevenkuhn, line 61: Is there going to be an issue setting the default package if there is an issue downloading that package?

ctacke: Agreed.  I think the edge case bryan was having was "default is an older version.  I told it to download the latest.  I already had the latest, so it didn't set that as the default" where normally downloading latest sets it as the default.

stevenkuhn: In the `if (collection[Version] != null && Force == false)` block starting on line 78, should the default package be set there? Effectively, if the latest version _is_ already available then go ahead and set the default package if it hasn't been updated already?

adrianstevens: Yeah can do - depends on what you think the purpose of latest.txt is  ... should is record the latest version available or record the latest version on the dev system?

Either way we have to handle errors but if you have a strong pref I can change it.

stevenkuhn (approved):

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareDownloadCommand.cs
adrianstevens: ahh perfect - and fixed

ctacke (approved):

PR #467, opened 2024-02-13
Improve firmware write behavior

adrianstevens:


Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareWriteCommand.cs
```diff
@@ -121,7 +121,6 @@ protected override async ValueTask ExecuteCommand()
 
         if (package == null)
         {
-            Logger?.LogError($"Firmware write failed - No package selected");
             return;
```
ctacke, line 124: This should throw a CommandException, otherwise the base command will return 0

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareWriteCommand.cs
```diff
@@ -291,11 +292,9 @@ protected override async ValueTask ExecuteCommand()
...
     private async Task WriteEspFiles(IMeadowConnection? connection, DeviceInfo? deviceInfo, FirmwarePackage package)
     {
-        if (connection == null)
-        {
-            connection = await GetConnectionAndDisableRuntime();
-            if (connection == null) return; // couldn't find a connected device
-        }
+        connection ??= await GetConnectionAndDisableRuntime();
+
+        if (connection == null) { return; } // couldn't find a connected device
```
ctacke, line 297: ditto

Source/v2/Meadow.Cli/Commands/Current/Firmware/FirmwareWriteCommand.cs
adrianstevens: sounds good to me

adrianstevens: but let's do it as part of another work item

ctacke (approved):

ctacke (approved):

PR #484, opened 2024-02-18
Fix delete all and file list

adrianstevens:


Source/v2/Meadow.Cli/Commands/Current/File/FileDeleteCommand.cs
```diff
@@ -9,6 +10,8 @@ public class FileDeleteCommand : BaseDeviceCommand<FileDeleteCommand>
     [CommandParameter(0, Name = "MeadowFile", IsRequired = true)]
     public string MeadowFile { get; init; } = default!;
 
+    private const string MeadowRootFolder = "meadow0";
```
jorgedevs, line 13: Minor thing, but since is private, shouldnt it be lowercase?

jorgedevs (approved):

Source/v2/Meadow.Cli/Commands/Current/File/FileDeleteCommand.cs
adrianstevens: its a bit of a convention for static placeholders

PR #487, opened 2024-02-19
LogError when build fails

doingnz:
Add minimum LogError messages to inform user when 'app run' command exits. 

Specifically needed when the project fails to build as the Meadow.CLI simple exits to the command prompt without giving any indication why it stopped. Added others for completeness.

### DIFF
--- a/Source/v2/Meadow.Cli/Commands/Current/App/AppRunCommand.cs
+++ b/Source/v2/Meadow.Cli/Commands/Current/App/AppRunCommand.cs
@@ -63,11 +63,13 @@ public class AppRunCommand : BaseDeviceCommand<AppRunCommand>
 
         if (!await BuildApplication(path, CancellationToken))
         {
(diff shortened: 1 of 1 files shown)


doingnz:
consider adding a custom MSBuild logger that can report progress and errors during the build process as described here: https://learn.microsoft.com/en-gb/previous-versions/visualstudio/visual-studio-2015/msbuild/build-loggers?view=vs-2015&redirectedfrom=MSDN

I expect it will be straightforward to adapt either example MS provides and replace the Console output with Meadow Logger calls to keep the logging uniform with remaining Meadow.CLI.

adrianstevens:
thx @doingnz  - I'm going to approach this a slightly different way but we'll fix asap

adrianstevens:
Thanks again @doingnz - we're now throwing the correct exceptions which will also give the correct user feedback. I'm closing this PR but really appreciate the nudge to get this updated.

PR #488, opened 2024-02-19
Add profiler CLI command

duduita:
- https://github.com/WildernessLabs/Meadow_Issues/issues/438
- https://github.com/WildernessLabs/Meadow/pull/566

**Summary**
- Only write the profiling data after a header is found, to avoid corrupted files due to the potential residual data in the UART serial
- If another header is found, start to write a new file, to avoid corrupted files as well (e.g. device resetting while profiling data)


Source/v2/Meadow.Hcom/SerialRequests/RequestType.cs
```diff
@@ -36,7 +36,8 @@ public enum RequestType : ushort
     HCOM_MDOW_REQUEST_MONO_FLASH = 0x19 | ProtocolType.HCOM_PROTOCOL_HEADER_ONLY_TYPE,
     HCOM_MDOW_REQUEST_SEND_TRACE_TO_UART = 0x1a | ProtocolType.HCOM_PROTOCOL_HEADER_ONLY_TYPE,
     HCOM_MDOW_REQUEST_NO_TRACE_TO_UART = 0x1b | ProtocolType.HCOM_PROTOCOL_HEADER_ONLY_TYPE,
-
+    HCOM_MDOW_REQUEST_SEND_PROFILER_TO_UART = 0x23 | ProtocolType.HCOM_PROTOCOL_HEADER_ONLY_TYPE,
```
duduita, line 39: It was not clear to me which value I should use, so I tried to use the lowest values available. Anyway, let me know if I should use other values:
```
HCOM_MDOW_REQUEST_SEND_PROFILER_TO_UART = 0x23
HCOM_MDOW_REQUEST_NO_PROFILER_TO_UART = 0x24
```

adrianstevens (approved):

PR #498, opened 2024-02-22
CLI changes required to at least build for VS2022

CartBlanche:


Source/v2/Meadow.SoftwareManager/F7FirmwareDownloadManager.cs
```diff
@@ -1,4 +1,7 @@
-﻿namespace Meadow.Software;
+﻿using Meadow.Cloud.Client;
```
adrianstevens, line 1: Are these usings needed? I don't see any code changes that would require them

Source/v2/Meadow.Hcom/Meadow.HCom.csproj
```diff
@@ -4,7 +4,9 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
-	<LangVersion>10</LangVersion>
+	  <LangVersion>10</LangVersion>
+    <SignAssembly>true</SignAssembly>
```
adrianstevens, line 8: interesting - I assume this is required for the extension?

CartBlanche: Yes. Assemblies need to be signed.

Source/v2/Meadow.Hcom/AssemblyAttributes.cs
```diff
@@ -1,3 +1,3 @@
 ﻿using System.Runtime.CompilerServices;
 
-[assembly: InternalsVisibleTo("Meadow.HCom.Integration.Tests")]
\ No newline at end of file
+// TODO [assembly: InternalsVisibleTo("Meadow.HCom.Integration.Tests")]
```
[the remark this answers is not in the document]
CartBlanche: Yes it could be, but then the unit test assembly also needs to be signed.

Source/v2/Meadow.Cli/SettingsManager.cs
```diff
@@ -1,5 +1,4 @@
 ﻿using System.Configuration;
-using System.Text.Json;
```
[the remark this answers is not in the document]
CartBlanche: I think it is, some of the usings cleanup got messed up. I think it's fixed now.

adrianstevens:
Looking at the code changes and the changes for the VS2022 extension ..... we should be close to being able to use the projects in the Meadow.CLI project directly without the need to do any file linking. We also shouldn't need any of the logic changes. The one thing missing is the code signing.

PR #499, opened 2024-02-23
Mac deploy output hack

adrianstevens:


CartBlanche (approved):
Not sure this will pass muster. ;)

PR #504, opened 2024-02-27
Small fix to make room for file progress output

adrianstevens:


Source/v2/Meadow.Cli/Commands/Current/File/FileWriteCommand.cs
```diff
@@ -69,6 +69,9 @@ protected override async ValueTask ExecuteCommand()
                 }
             }
         }
+
+        //add a black line after writing the file write progress
+        Logger?.LogInformation(string.Empty);
```
stevenkuhn, line 74: Small nit: `?` is no longer required.

stevenkuhn (approved):

Source/v2/Meadow.Cli/Commands/Current/File/FileWriteCommand.cs
adrianstevens: Let's get all of these PRs merged and then we can sweep the project

PR #536, opened 2024-03-15
Remove auth from firmware download, minor defensive coding, version bump

adrianstevens:


Source/v2/Meadow.Tooling.Core/Package/PackageManager.cs
```diff
@@ -152,13 +152,16 @@ public bool BuildApplication(string projectFilePath, string configuration = "Rel
...
+            if (opts != null && opts.Deploy != null)
             {
-                noLink = opts.Deploy.NoLink;
-            }
-            if (opts.Deploy.IncludePDBs != null)
-            {
-                includePdbs = opts.Deploy.IncludePDBs.Value;
+                if (opts.Deploy.NoLink != null && opts.Deploy.NoLink.Count > 0)
+                {
+                    noLink = opts.Deploy.NoLink;
```
ctacke, line 159: Maybe log out the nolink items here?

ctacke (approved):

Source/v2/Meadow.Tooling.Core/Package/PackageManager.cs
adrianstevens: let's do it

PR #539, opened 2024-03-19
Better messaging if runtime enable/disable fails

adrianstevens:


Source/v2/Meadow.Cli/Commands/Current/Runtime/RuntimeDisableCommand.cs
```diff
@@ -28,6 +28,11 @@ protected override async ValueTask ExecuteCommand()
 
             state = await device.IsRuntimeEnabled(CancellationToken);
 
+            if (state == true)
+            {
```
ctacke, line 32: Should throw a CommandException with this text

Source/v2/Meadow.Cli/Commands/Current/Runtime/RuntimeEnableCommand.cs
```diff
@@ -28,6 +28,11 @@ protected override async ValueTask ExecuteCommand()
 
             state = await device.IsRuntimeEnabled(CancellationToken);
 
+            if (state == false)
+            {
+                Logger?.LogError("Failed to enable runtime");
```
ctacke: ibid

Source/v2/Meadow.Cli/Commands/Current/Runtime/RuntimeDisableCommand.cs
adrianstevens: Good call - updated

ctacke (approved):

Source/v2/Meadow.Cli/Commands/Current/Runtime/RuntimeEnableCommand.cs
adrianstevens: done

PR #545, opened 2024-03-27
bug fix for reading large files. bug fix for no local file name

ctacke:


Source/v2/Meadow.Hcom/SerialResponses/SerialResponse.cs
```diff
@@ -18,8 +18,13 @@ internal class SerialResponse
     public uint UserData => BitConverter.ToUInt32(_data, HCOM_PROTOCOL_REQUEST_HEADER_USER_DATA_OFFSET);
     protected int PayloadLength => _data.Length - RESPONSE_PAYLOAD_OFFSET;
 
-    public static SerialResponse Parse(byte[] data, int length)
+    public static SerialResponse? Parse(byte[] data, int length)
     {
+        if (length == 0)
+        {
+            return null;
```
alexischr, line 25: Can we avoid the nulls and return an empty SerialResponse? It can be a static readonly `EmptySerialResponse` if object churn is a concern (this is the CLI, so probably not)

adrianstevens (approved):

PR #548, opened 2024-04-07
Ensure we re-enable the Meadow regardless of what happens during the …

CartBlanche:
…deploy process.

Fixes: https://github.com/WildernessLabs/Meadow_Issues/issues/585


Source/v2/Meadow.Cli/Commands/Current/App/AppDeployCommand.cs
```diff
@@ -31,8 +31,6 @@ protected override async ValueTask ExecuteCommand()
 
         var connection = await GetCurrentConnection();
 
-        await AppTools.DisableRuntimeIfEnabled(connection, Logger, CancellationToken);
```
CartBlanche, line 34: Should AppTools and Strings be moved into Tooling.Core?

CartBlanche: @adrianstevens, @ctacke : I've move AppTools, Strings. etc to Tooling.Core as Tooling Core is shared. Thoughts?

Meadow.CLI.Core/Devices/MeadowDeviceHelper.cs
```diff
@@ -252,16 +252,17 @@ public async Task DeployApp(string fileName, bool includePdbs = false, Cancellat
...
-                await MonoEnable(true, cancellationToken: cancellationToken);
-
-                await Task.Delay(2000, cancellationToken);
             }
             catch (Exception ex)
             {
-                await MonoDisable(true, cancellationToken);
                 throw ex;
             }
+            finally
```
adrianstevens, line 260: Do we need this in the finally block?

Source/v2/Meadow.Cli/Commands/Current/App/AppDeployCommand.cs
adrianstevens: the Strings are for the CLI command line app, my gut is to have resource files that are app-specific but we can talk about it tomorrow

adrianstevens: Also why did this line get removed?

Source/v2/Meadow.Tooling.Core/Strings.cs
ctacke, line 1: This is very CLI-specific and doesn't need to be in a shared library

Source/v2/Meadow.Cli/Commands/Current/App/AppDeployCommand.cs
CartBlanche: @adrianstevens it was moved. I mentioned a couple of calls ago, about centralising the enablement and disablement of around App Deployment.

Meadow.CLI.Core/Devices/MeadowDeviceHelper.cs
CartBlanche: @adrianstevens If an exception is thrown, we still want it to try to re-enable the runtime. Hence why it was moved into the finally block. The original issue Chris raised mentions that.

Source/v2/Meadow.Tooling.Core/Strings.cs
CartBlanche: @ctacke As per a previous comment on this PR, if I don't move it there, then the string literals required would be duplicated. I'm all for have a Meadow.CLI.Localisation or similar solution, that centralises all of our strings. That will make it easier to support other languages, which we spoke about some time ago, on a call.

adrianstevens: That's because code that is specific to the CLI has been moved to Core. I personally don't support these changes.

adrianstevens:
I think this PR has strayed away from the core the issue it's trying to solve: https://github.com/WildernessLabs/Meadow_Issues/issues/585

This should simply be a matter of enabling Meadow from within the extension project after deployment.

PR #549, opened 2024-04-08
Add telemetry support

stevenkuhn:
This adds support for collecting anonymous usage data to help understand how the CLI is used and how to help improve the Meadow CLI, Workbench, and IDE extensions.

## How to opt in

The Meadow CLI displays a message similar to the one below when you first run a Meadow CLI command and have not yet decided to participate or not.

```
Let's improve the Meadow experience together
--------------------------------------------
To help improve the Meadow CLI, we'd like to collect anonymous usage data. This data helps
us understand how the CLI is used, so we can make it better for everyone. This usage data
not tied to individuals and no personally identifiable information is collected.

Our privacy policy is available at https://www.wildernesslabs.co/privacy-policy.

You can change your mind at any time by running the "meadow telemetry [enable|disable]"
command or by setting the MEADOW_TELEMETRY environment variable to '1' or '0' ('true' or
'false', respectively).

Would you like to participate? [y/n] (y):
```

Once a user has given their choice, that choice is saved to the Meadow CLI configuration (that can be viewed with `meadow config --list`) and this message will no longer appear.

This can be overridden entirely by setting the `MEADOW_TELEMETRY` environment variable to `1`/`true` to opt-in or `0`/`false` (to opt-out). If that environment variable is set, then this message will not be displayed.

A best effort is made to determine if the Meadow CLI command is being run in a continuous integration (CI) environment. If it is, then telemetry is **opt-out and disabled by default**.

## Data points

The usage collection is for anonymous usage data only. It doesn't collect personal data such as username or email address, nor anything about the command being run outside of the command/subcommand name (no command arguments are collected).

This feature collects the following information:

_The examples are based on a running the command `meadow cloud command publish my_command --collectionId 12345`._

| Data | Example | Notes |
|-|-|-|
| Command | `cloud command publish` | |
| SessionId | `hDtIQEoBKMbicpggF4r+wIB61Fjw68aP` | This is a random string generated at runtime and used only while the command is executed. |
| MachineId | `bb4d522020164e4cadec605c4b0521d4` | This is a random Guid that is generated at runtime once and stored in the Meadow CLI config. If this value is found in the config, that value is used. |
| Assembly Name | `meadow` | |
| Assembly Version| `2.0.33` | |
| OS Version | `10.0.22631` | |
| OS Platform | `Windows` | |
| OS Architecture | `X64` | |
| Kernel Version | `Microsoft Windows 10.0.22631` | |
| Runtime Identifier | `win-x64` | |

### DIFF
--- a/Source/v2/Meadow.Cli/Commands/Current/BaseCommand.cs
+++ b/Source/v2/Meadow.Cli/Commands/Current/BaseCommand.cs
@@ -1,6 +1,10 @@
 ﻿using CliFx;
+using CliFx.Attributes;
 using CliFx.Infrastructure;
+using Meadow.Telemetry;
 using Microsoft.Extensions.Logging;
+using Spectre.Console;
+using System.Reflection;
 
 namespace Meadow.CLI.Commands.DeviceManagement;
 
@@ -28,6 +32,23 @@ public abstract class BaseCommand<T> : ICommand
             _console = console;
             CancellationToken = _console.RegisterCancellationHandler();
 
+            try
+            {
+                if (MeadowTelemetry.Current.ShouldAskForConsent)
+                {
+                    AnsiConsole.MarkupLine(Strings.Telemetry.ConsentMessage);
+
+                    var result = AnsiConsole.Confirm(Strings.Telemetry.AskToParticipate, defaultValue: true);
+                    MeadowTelemetry.Current.SetTelemetryEnabled(result);
+                }
+
+                MeadowTelemetry.Current.TrackCommand(GetCommandName());
+            }
+            catch
+            {
+                // Swallow any telemetry-related exceptions
+            }
+
             await ExecuteCommand();
         }
         catch (Exception ex) when (ex is not CommandException && ex is not CliFx.Exceptions.CommandException)
@@ -39,5 +60,10 @@ public abstract class BaseCommand<T> : ICommand
         {
             throw new CommandException("Cancelled", CommandExitCode.UserCancelled);
         }
+    }
+
+    private string? GetCommandName()
+    {
+        return GetType().GetCustomAttribute<CommandAttribute>(true)?.Name;
     }
 }
--- a/Source/v2/Meadow.Cli/Meadow.CLI.csproj
+++ b/Source/v2/Meadow.Cli/Meadow.CLI.csproj
@@ -30,15 +30,16 @@
   </PropertyGroup>
 
   <ItemGroup>
-  	<PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="7.0.0" />
-  	<PackageReference Include="Serilog" Version="3.0.1" />
-  	<PackageReference Include="CliFx" Version="*" />
-  	<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
-  	<PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
-  	<PackageReference Include="System.Configuration.ConfigurationManager" Version="7.0.0" />
-  	<PackageReference Include="System.Management" Version="7.0.2" />
-  	<PackageReference Include="Serilog.Sinks.Console" Version="4.1.0" />
-  	<PackageReference Include="Serilog.Extensions.Logging" Version="7.0.0" />
+    <PackageReference Include="CliFx" Version="*" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="7.0.0" />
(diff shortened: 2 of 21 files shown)


stevenkuhn:
@adrianstevens @ctacke I updated the NuGet packages for the solution so I need to validate everything is still good to go for the extensions.

I added the [Spectre.Console](https://spectreconsole.net/) library to handle the confirmation prompt.

Source/v2/Meadow.Cli/Commands/Current/BaseCommand.cs
```diff
@@ -28,6 +32,23 @@ public async ValueTask ExecuteAsync(IConsole console)
             _console = console;
             CancellationToken = _console.RegisterCancellationHandler();
 
+            if (MeadowTelemetry.Current.ShouldAskForConsent)
+            {
+                AnsiConsole.MarkupLine(@$"
```
adrianstevens, line 37: would love to move this text into strings.cs

adrianstevens (approved):

Source/v2/Meadow.Cli/Commands/Current/BaseCommand.cs
stevenkuhn: Good idea! I will move the text over there.

Source/v2/Meadow.Cli/Meadow.CLI.csproj
```diff
@@ -30,15 +30,17 @@
...
 		<PackageReference Include="CliFx" Version="*" />
-		<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="7.0.0" />
-		<PackageReference Include="Microsoft.Extensions.Http" Version="7.0.0" />
-		<PackageReference Include="System.Configuration.ConfigurationManager" Version="7.0.0" />
-		<PackageReference Include="System.Management" Version="7.0.2" />
-		<PackageReference Include="Serilog.Sinks.Console" Version="4.1.0" />
-		<PackageReference Include="Serilog.Extensions.Logging" Version="7.0.0" />
+		<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="8.0.0" />
+		<PackageReference Include="Microsoft.Extensions.Http" Version="8.0.0" />
+		<PackageReference Include="Spectre.Console" Version="0.48.0" />
```
adrianstevens, line 39: Is this being used?

Source/v2/Meadow.Tooling.Core/Meadow.Tooling.Core.csproj
```diff
@@ -8,9 +8,10 @@
 
   <ItemGroup>
 	<PackageReference Include="Glob" Version="1.1.9" />
+	<PackageReference Include="Microsoft.ApplicationInsights" Version="2.22.0" />
+    <PackageReference Include="System.Configuration.ConfigurationManager" Version="8.0.0" />
```
adrianstevens, line 12: alignment :)

Source/v2/Meadow.Cli/Meadow.CLI.csproj
stevenkuhn: Are you asking about `Spectre.Console`? Yes, it's being used for the confirmation prompt and to give some highlighting to a part of the message: 

```csharp
AnsiConsole.MarkupLine(Strings.Telemetry.ConsentMessage);
var result = AnsiConsole.Confirm(Strings.Telemetry.AskToParticipate, defaultValue: true);
```

image

Source/v2/Meadow.Tooling.Core/Meadow.Tooling.Core.csproj
stevenkuhn: Since there's no `.editorconfig` file in this repo, what is the standard for tabs/spaces in .csproj files? Is it tabs? Or spaces? If spaces, 2 spaces or 4? The default for VS2022 is 4 spaces for each indent. If you want that, I can update all the .csproj files.

stevenkuhn: I see that quite a few `.csproj` files use 2 spaces for indents. I'll add an `.editorconfig` and standardize on that.

adrianstevens: Oh amazing - thank you

Source/v2/Meadow.Cli/Meadow.CLI.csproj
adrianstevens: Ok excellent - in that case I might use it somewhere else in the app

PR #551, opened 2024-04-18
Validate firmware before running or trimming + minor cleanup

adrianstevens:


Source/v2/Meadow.Cli/Commands/Current/App/AppRunCommand.cs
```diff
@@ -23,14 +24,33 @@ public class AppRunCommand : BaseDeviceCommand<AppRunCommand>
     [CommandOption("nolink", Description = Strings.NoLinkAssemblies, IsRequired = false)]
     public string[]? NoLink { get; private set; }
 
-    public AppRunCommand(IPackageManager packageManager, MeadowConnectionManager connectionManager, ILoggerFactory loggerFactory)
+    readonly FileManager _fileManager;
```
ctacke, line 27: Underscore??? :P

ctacke (approved):

Source/v2/Meadow.Cli/Commands/Current/App/AppRunCommand.cs
adrianstevens: well I think _someone_ used them when they created the CLI v2 project so I'm running with it ;)

PR #557, opened 2024-05-01
Restore console output for CLI only when writing firmware via DFU

adrianstevens:


Source/v2/Meadow.Dfu/DfuUtils.cs
```diff
@@ -58,11 +58,9 @@ private static void FormatDfuOutput(string logLine, ILogger? logger, DfuFlashFor
         }
         else //Console out
         {
-            logger?.LogInformation(logLine);
```
ctacke, line 61: NYET!

ctacke (approved):

PR #560, opened 2024-05-04
project new command

CartBlanche:


Source/v2/Meadow.CLI/Commands/Current/App/AppTools.cs
```diff
@@ -133,4 +134,32 @@ internal static string SanitizeMeadowFilename(string fileName)
 
         return meadowFileName!.Replace(Path.DirectorySeparatorChar, '/');
     }
+
+    // Method to run a command and return the output as a list of strings
+    public async static Task<List<string>> RunDotNetCommand(string command, string arguments, CancellationToken cancellationToken)
```
CartBlanche, line 139: @adrianstevens, @jorgedevs, @ctacke  I've put this in here for now. Thoughts on what's your preferred way to have a centralised simple `Process` command, that we can re-use?

Source/v2/Meadow.Cli/Strings.cs
```diff
@@ -71,19 +72,46 @@ public static class Strings
...
+        public const string Installed = "The following Meadow Project Templates are now installed.";
+
+        public const string WizardTitle = "Meadow Project Wizard...";
+        public const string ProjectName = "What is your project's name?";
+        public const string GenerationComplete = "All of {0}'s projects have now been created!";
+        public const string GenerateSln = "Create an sln that contains all the new projects?";
+        public const string NoTemplatesFound = "No Meadow project templates found.";
+        public const string InstalledTemplates = "--- Installed Templates ---";
+        public const string MoreChoicesInstructions = "(Move up and down to reveal more templates)";
+        public const string Instructions = "Press {0} to toggle a template or {1} to accept and generate the selected templates)";
```
stevenkuhn, line 101: Does this need a beginning parenthesis or the end one removed?

CartBlanche: Nicely spotted. Removed.

ctacke:
I'm also in the camp of "this is not a good idea".  If we're simply trying to make it easier a user who forgets the command, why not just echo out the proper command for them?  Creating a new command that creates a new process and calls some other existing CLI seems clunky AF

adrianstevens:
I'll put some time into this one today and get it merged - thx for the patience Dom :)

adrianstevens (approved):

adrianstevens (approved):

PR #563, opened 2024-05-15
Provisioning of multiple attached devices command

CartBlanche:
Current process:
1. The `meadow provision` command is run on the command line and all devices in bootloader mode will be listed:
image

2. Selected which devices you would like to provision:
image

3. Confirm you want to proceed:
image

4. The specific firmware will be downloaded and each device will be flashed and updated sequentially:
image

5. If all goes well, all devices will have completed successfully
image

If a provision.json file is included in the same directory, where the command is run, the provision.json file should look something like this:

```
{
    "AppPath": "C:\\Users\\willo\\Downloads\\Blinky\\Debug\\netstandard2.1\\App.dll",
    "Configuration": "Debug",
    "DeployApp": "false",
    "FirmwareVersion": "1.12.2.0"
}
```
It will then use these settings for all attached devices and it will attempt to Deploy the App.dll, after trimming, from the path mentioned in "AppPath"

Tested on:
- [x] Mac
- [x] Linux
- [x] Windows

TODO post `v1.x` Release: 
Get this working in parallel. Currently there seems to be a conflict after the DFU process (Multi-OS flashing works fine) has finished and more than 1 device is connected, so the RunTime and ESP steps fail.


Source/v2/Meadow.CLI/Commands/Current/Provision/ProvisionCommand.cs
```diff
@@ -0,0 +1,234 @@
...
+        else
+            osVersion = OsVersion;
+
+        // Install DFU, if it's not already installed.
+        var dfuInstallCommand = new DfuInstallCommand(SettingsManager, LoggerFactory);
+        await dfuInstallCommand.ExecuteAsync(Console);
+
+        // Make sure we've downloaded the passed in osVersion or default
+        // Use Firmware Download command??
+        if (await MeadowCLI($"firmware download -v {osVersion} -f") == 0)
```
ctacke, line 50: Am I reading this right?  The CLI is shelling out to create another instance of the CLI instead of just using the APIs it has access to?

CartBlanche: @ctacke As discussed on that call I'll be changing these shelling calls to use the API.

ctacke:
I like the tool, but I'm not certain it should be part of the main CLI - it feels like a separate tool to me.

PR #564, opened 2024-05-16
Try to compare path parts to get correct relative path and avoid star…

CartBlanche:
…tIndex error.


adrianstevens:
This will break on Windows - closing PR

PR #578, opened 2024-06-12
Fix profiler output file refs

duduita:
Some `refs` were removed in this [PR](https://github.com/WildernessLabs/Meadow.CLI/commit/a38f3cfe5d3da98f8f6db84f8f4bd74adc8a89cd), which is causing issues when using the profiler.

### DIFF
--- a/Source/v2/Meadow.CLI/Commands/Current/Uart/UartProfilerEnableCommand.cs
+++ b/Source/v2/Meadow.CLI/Commands/Current/Uart/UartProfilerEnableCommand.cs
@@ -18,7 +18,7 @@ public class UartProfilerEnableCommand : BaseDeviceCommand<UartProfilerEnableCom
         : base(connectionManager, loggerFactory)
     { }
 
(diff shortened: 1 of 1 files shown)


Source/v2/Meadow.CLI/Commands/Current/Uart/UartProfilerEnableCommand.cs
```diff
@@ -18,7 +18,7 @@ public UartProfilerEnableCommand(MeadowConnectionManager connectionManager, ILog
         : base(connectionManager, loggerFactory)
     { }
 
-    private void StartNewFile(string outputPath, FileStream? outputFile, byte[] header, ref int headerIndex, ref int totalBytesWritten, ref int headerFileCount)
+    private void StartNewFile(string outputPath, ref FileStream? outputFile, byte[] header, ref int headerIndex, ref int totalBytesWritten, ref int headerFileCount)
```
adrianstevens, line 21: FileStream is a class - e.g. a reference type, not a value type so I don't believe this `ref` is necessary

adrianstevens (approved):

PR #581, opened 2024-06-25
Add Mac and Linux support to dfu install.

CartBlanche:
Refactor to centralise Process launching and use more modern async pattern.

Changes:
Mac - will new install brew, then install dfu-util if they are not already installed.
Linux - Currently only Ubuntu is supported, but we can add more support for their respective package managers in future.  It prompts for `sudo` rights  before calling `apt-get`.
Windows - Will now auto-prompt for UAC access, before running the commands, so no need to launch a new prompt with Admin rights (I don't think, might need more testing, to be 100% sure)

Tested on:
- [x] Mac
- [x] Linux
- [x] Windows

Fixes https://github.com/WildernessLabs/Meadow_Issues/issues/733


Source/v2/Meadow.Dfu/DfuUtils.cs
```diff
@@ -301,4 +277,301 @@ private static string GetDfuUtilVersion()
...
+
+                    // We check the version again, because on some versions of Ubuntu the default dfu-util version is 0.9 :(
+                    var installedDfuUtilVersion = new Version(await GetDfuUtilVersion());
+                    var expectedDfuUtilVersion = new Version(DEFAULT_DFU_VERSION);
+                    if (installedDfuUtilVersion.CompareTo(expectedDfuUtilVersion) < 0)
+                    {
+                        var dfuPackageUrl = RuntimeInformation.OSArchitecture switch
+                        {
+                            Architecture.Arm64 => DFU_UTIL_UBUNTU_ARM64_URL,
+                            Architecture.X64 => DFU_UTIL_UBUNTU_AMD64_URL,
```
halyssonJr, line 455: We should consider including the AMD architecture. The binaries work for x86 and AMD

halyssonJr (approved):
LGTM :)

ctacke:
I'm not sure I'm a fan of this whole strategy.  We're basically trying to wrap a CLI command with a CLI command, and if the versions of DFU change or whatever, then we have to update as well.  An alternative would be that if DFU is missing, give the user an error that is meaningful, for example:

```
> meadow firmware update
ERROR: DFU Is not found.
DFU is required to write the firmware.  For more information see the docs at wildernesslabs.co/docs?id=12345
```

bryancostanich:
This one is also my fault. We added windows DFU install support ages back because customers were running into serious issues with it. This path was fine for windows, but with mac and linux, this solution is super ugly. Hat tip to @CartBlanche for tracking this down and trying to solve it, but this is winding up being a super brittle path forward and will ultimately be superseded when the OtA path for firmware updates is moved to in the CLI, which will relegate DFU to only initial firmware flashing and hard erase. And initial firmware flashing will largely be handled in the manufacturing/fulfillment pipeline anyway.

PR #590, opened 2024-08-20
Re-order Debugging Start-up to be like v1. Refactor MeadowConnectionManager so it can be shared amongst extensions

adrianstevens:


Source/v2/Meadow.Tooling.Core/Connection/MeadowConnectionManager.cs
```diff
@@ -35,22 +37,36 @@ public MeadowConnectionManager(ISettingsManager settingsManager)
             throw new Exception($"No 'route' configuration set.{Environment.NewLine}Use the `meadow config route` command. For example:{Environment.NewLine}  > meadow config route COM5");
         }
 
-        return GetConnectionForRoute(route, forceReconnect);
+        return await GetConnectionForRoute(route, forceReconnect);
     }
 
-    public IMeadowConnection? GetConnectionForRoute(string route, bool forceReconnect = false)
+    public static async Task<IMeadowConnection?> GetConnectionForRoute(string route, bool forceReconnect = false)
```
adrianstevens, line 43: I'm not a fan of changing this to async unless we make further changes.

This method blocks and is potentially long running because of the loop. The async is only supporting the change from Thread.Sleep to Task.Run.

adrianstevens: I am a fan of the rest of the defensive coding

Source/v2/Meadow.Hcom/Connections/SerialConnection.ListenerProc.cs
```diff
@@ -1,354 +0,0 @@
-﻿namespace Meadow.Hcom
```
adrianstevens, line 1: Is this not needed at all?

Source/v2/Meadow.HCom/Connections/ConnectionBase.cs
```diff
@@ -64,11 +64,11 @@ protected void RaiseConnectionMessage(string message)
 
     bool firstRun = true;
 
-    protected void RaiseDebuggerMessage(byte[] data)
+    protected async Task RaiseDebuggerMessage(byte[] data)
     {
         if (firstRun)
         {
-            Thread.Sleep(2000);
+            await Task.Delay(2000);
```
ctacke, line 71: Is there a specific need for this change?  Was Thread.Sleep not working?

Source/v2/Meadow.HCom/Connections/SerialConnection.ListenerProc.cs
ctacke, line 1: Why was this deleted wholesale?  Was it dead code that was not used, or did the functionality get moved somewhere?

Source/v2/Meadow.Hcom/Connections/SerialConnection.ListenerProc.cs
CartBlanche: @adrianstevens if you look at the path, this is referring to `Hcom` and not `HCom`, so it's actually a rename, and the file does exists under the `HCom` path here - https://github.com/WildernessLabs/Meadow.CLI/blob/e0d60798920b4d01da7219ab7535ac1403690b01/Source/v2/Meadow.HCom/Connections/SerialConnection.ListenerProc.cs

Source/v2/Meadow.HCom/Connections/SerialConnection.ListenerProc.cs
CartBlanche: It wasn't it related to the HCom case sensitivity move. Everything builds with the new path and thaat file exists there.

adrianstevens:
This is out of sync now with Alexis' latest changes - closing PR

PR #643, opened 2025-09-21
Create missing local folders during a file read (fixes hanging bug)

adrianstevens:


Source/Meadow.HCom/Connections/SerialConnection.cs
```diff
@@ -1134,12 +1134,10 @@ public override async Task<bool> ReadFile(string meadowFileName, string? localFi
         command.MeadowFileName = meadowFileName;
         command.LocalFileName = localFileName;
 
-        var completed = false;
         Exception? ex = null;
 
         void OnFileReadCompleted(object? sender, string filename)
```
CartBlanche: So this method is essentially empty now, correct?

adrianstevens: fixed - thx!

